Issue #7, opened 2022-03-21
Calculate disk usage of a model


miquelduranfrigola:
# Background
When a model is downloaded with the `fetch` command, the corresponding GitHub repository is cloned, including Git LFS files. As of now, there is no way to know the size of the downloaded files before cloning. In addition, a conda environment or a docker container is created after download. We do not have a way of estimating the size of such environments/containers. Therefore, we are unable to inform the user about the size of the model they are about to fetch.

# Requested feature
We would like to calculate total disk usage of a model **before** fetching it. This includes:
* Size of files from GitHub (including Git LFS)
* Size of conda environment or docker container

Riyabelle25:
Hey there! I can take this issue up

GemmaTuron:
Hello @Riyabelle25,

This is great thanks. To tackle this issue can you please first answer in issue #36 ?

Riyabelle25:
@GemmaTuron thanks!
I've made a PR to fix this issue first, will then move to the other issues I showed an interest in (if they're still open then)

miquelduranfrigola:
Hi @Riyabelle25 please refer to issue #6. If you want to contribute I will be happy to add you as an assignee! Many thanks for your amazing contributions. Closing issue now.

PR #38, opened 2022-03-28
[add] Calculate disk usage of fetched model and display the same

Riyabelle25:
Solves #7 
Using the very handy GitHub API to get information about the model repo in question, then grepping the API response to simply show the `size` in KB. 
My changes to achieve this are:

```bash
        import os

        url = "https://github.com/ersilia-os/{0}".format(model_id)
        cmd = "echo " + url + "| perl -ne 'print $1 if m!([^/]+/[^/]+?)(?:\.git)?$!' | xargs -I{} curl -s -k https://api.github.com/repos/'{}' | grep size"
        echo(
            "The disk storage of this model in KB is"
        )
        os.system(cmd) 
```

Attached herein is how the output looks currently:
Screenshot 2022-03-28 at 3 30 23 PM

### DIFF
--- a/ersilia/cli/commands/fetch.py
+++ b/ersilia/cli/commands/fetch.py
@@ -1,5 +1,6 @@
 import click
 
+import os
 from . import ersilia_cli
 from .. import echo
 from ...hub.fetch.fetch import ModelFetcher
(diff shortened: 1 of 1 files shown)


Riyabelle25:
Hi @miquelduranfrigola, do review my solution and let me know your thoughts.
This is a quick fix, as I have used the GitHub API before in previous open-source projects

miquelduranfrigola:
Hi @Riyabelle25 - many thanks for this good first step. Is there a way to track the size of files stored in Git LFS through the GitHub API? I like your solution but, since big files are stored in Git LFS, those would be the most important to track.

Riyabelle25:
GitHub API currently does not support tracking the size of the LFS files, though it is a [feature request](https://github.community/t/feature-request-can-github-api-show-repo-size-including-git-lfs/13991).

I have been exploring other solutions, and a tentative one would be:

1.  `cat` the `.gitattributes` to get the file extension that would be tracked as LFS. 
2. Iterate through the remote repo, find the files ending with that extension
3. `grep` through each file's content, for their listed `size`. (I read that pointer files have the size of the LFS file mentioned in them)

miquelduranfrigola (approved):
I am merging this PR (thanks!) as we still try to fix the Git LFS issue.

Riyabelle25:
Thank you @miquelduranfrigola!

PR #55, opened 2022-03-30
Correct grammar in the Readme file under the Road map section

kbetty:
Have created this PR as part of improving our documentation in the Read me file

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ The Ersilia Model Hub is developed and maintained by a small team of Ersilia emp
 The Ersilia Open Source Initiative adheres to the [Contributor Covenant](https://ersilia.gitbook.io/ersilia-wiki/code-of-conduct) guidelines.
 
 # Roadmap
-We are working to grow the Hub organically and responding to our users needs. Here a detail of the next features to come, stay tuned!
+We are working to grow the Hub organically and responding to our users needs. Here are the details of the next features to come, stay tuned!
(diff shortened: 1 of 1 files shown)


GemmaTuron (approved):
Hi @kbetty 

Thanks for the change. Since you are working on it intensively, I ask you:
There is a major thing that is mistaken in our readme file. Can you spot what it is and correct it?

kbetty:
> Thanks for the change. Since you are working on it intensively, I ask you:
There is a major thing that is mistaken in our readme file. Can you spot what it is and correct it?

Welcome @GemmaTuron! 

Yes as part of documentation, we need to have our docs well documented with clear information to the readers(user or contributors ).

Am happy to improve our documentation, will take on the assigned task and will share my findings, perhaps suggest changes accordingly through a commit.

kbetty:
@GemmaTuron have gone through the our Readme file carefully and Identified the following:

- For each section in the readme file, there is a link that takes the reader to back to the top of the page. This is repeated on every section of the readme file which I suggest we should get rid of. 
Screenshot from 2022-03-31 08-56-57

- Some words in the paragraphs are capitalized of which some of them are supposed to be lowercase based on orthography. Forex ample a statement like `Low and Middle Income Country institutions` should be written as `low and middle income country institutions `

Let me know if there is something am missing out so that I can fix together with those that have identified above. Thanks

PR #61, opened 2022-03-30
Updated README

victorabba:
Added guidelines for onboarding new contributors.


miquelduranfrigola:
Hi @victorabba I greatly appreciate your input here. But I would prefer that we do the contributions on the main README file. Would that be possible?

PR #64, opened 2022-03-30
Updated_Contribution.md

victorabba:
Made changes to 'Contribute'.


miquelduranfrigola:
Hi @victorabba I appreciate your input here. I would prefer that we do not duplicate files. Can you contribute to the contribution file directly?

PR #111, opened 2022-04-04
Added context with image, which will help new contributors.

Jaya3112:
I have added some more context along with images so that it can be helpful for new contributors. Please review it.


GemmaTuron (approved):
Thanks for the suggestions!

@sayantani11 can you consider these when merging with the CONTRIBUTION.md file?

Issue #44, opened 2022-03-29
Guidelines for onboarding of new contributors


sayantani11:
**Is your feature request related to a problem? Please describe.**
As a contributor for quite sometime now, I can tell from my experience the initial experience can be stressful. On the other hand, a lot of new contributors are sometimes shy to open up about their issues and problems to the project maintainers, or discuss about in the public communication channels.

**Describe the solution you'd like**
In this situation, it will be great if we could introduce a readme file, specifically designed for the new contributors with all the relevant links or resources. This makes the process of contribution easy and smooth for new onboarding contributors and also encourages more future contributions.

sayantani11:
Please assign me this issue. Would like to work on it.

GemmaTuron:
Hi @Sayantani11, 

Good initiative. Please do it by creating a CONTRIBUTION.md file in the root of the repository. I might add things in this file as well!

Rufaida98:
@sayantani11 nice initiative

Jaya3112:
For new contributors, it is quite difficult and stressful to contribute because they do not know how to approach any problem smoothly. They have to hustle a lot to solve any issue. So if we provide some images that can direct them in a correct then it will be very easy for them to do their contribution. I want to add more to the  CONTRIBUTION.md file which will help new contributors. Please assign me this issue. I would like to work on it.

loweyvana:
@sayantani11 nice initiative. If you don't mind i would like to work with you on this. Being a new contributor, I must say i had a difficult time getting familiar here and I, therefore, think I can be of help to those coming after me. I could
-add demo gifs and/or demo screenshots to make the process easier and smoother.

Rufaida98:
@sayantani11 can you make a folder and anyone ads file like I want to add resources file which contain helpful resources for contributors is it possible?

Jaya3112:
> For new contributors, it is quite difficult and stressful to contribute because they do not know how to approach any problem smoothly. They have to hustle a lot to solve any issue. So if we provide some images that can direct them in a correct then it will be very easy for them to do their contribution. I want to add more to the CONTRIBUTION.md file which will help new contributors. Please assign me this issue. I would like to work on it.

@sayantani11 can I work on this issue. I have mentioned my idea above.

miquelduranfrigola:
Thanks @sayantani11 for an awesome initiative. A CONTRIBUTION.md file has been created in #70

[...]

PR #118, opened 2022-04-04
Add files via upload

Rufaida98:
this an editable word file where anyone can add resources that could be helpful for contributors #44


miquelduranfrigola:
Nice initiative, @Rufaida98. Do you think we can have this in markdown format, instead of docx?

Rufaida98:
@miquelduranfrigola OK I'll edit it out 👍

PR #161, opened 2022-04-06
Create README_medhavi.md 

Med16-11:
Created a readme file as mentioned in the task. Mentors' feedback is required to carry on with other tasks.


GemmaTuron (changes requested):
The Readme you have created appears as empty, can you fill it in with what you think are the necessary elements and information it should have?

GemmaTuron (approved):

Issue #168, opened 2022-04-07
Check that Git LFS is correctly installed


miquelduranfrigola:
Some large models (e.g. chemprop-antibiotic) require Git LFS to be installed. If Git LFS is not installed, then the model is fetched but only pointers to the checkpoints are given. As a result, users get empty outputs when running predictions with the model.

I would suggest that we add a check in the main `__init__.py` file to check that Git LFS is installed. If it is not, raise an exception. Happy to listen to other suggestions too!

Please respond below if you want to work on this issue and I will assign it to you.

mahamtariq58:
Hi @miquelduranfrigola , I would love to try this issue....

Nawarrr:
I would love to work on this, as I faced this problem yesterday, assign me if it's possible @miquelduranfrigola

Nawarrr:
@miquelduranfrigola  I also think that a warning is a better option, as some models (or maybe in the future if it is managed to make the models smaller) might not need git lfs and they who use it might be aware of that.

miquelduranfrigola:
Hi all, thanks for your interest in this. I am assigning this to the three of you. Feel free to join efforts if you want, or simply PR your individual solutions/suggestions. Thanks!

miquelduranfrigola:
This is already being addressed in the PRs, so closing the issue. Great job.

PR #180, opened 2022-04-08
Checking if git lfs is installed

Nawarrr:
Following Issue #168
- **What was the problem?**
git-lfs not being installed caused issues while fetching big sized models like chemprop-antibiotic
- **How did I tackle the problem ?**
I've faced the problem yesterday, there is a couple of sides to it, it's if the user have a conda environment installed or he doesn't, 
the only way I found to check the conda environment was checking environment variables for variables with word "CONDA" in it, in case no conda environment was found, it will check if git-lfs is installed with python pip.

I don't have experience in writing tests for something like this, I would love to learn something like that.

### DIFF
--- a/ersilia/__init__.py
+++ b/ersilia/__init__.py
@@ -71,14 +71,16 @@ __all__ = ["__version__"]
 
 # Checking for git-lfs
 import subprocess
-
+# Incase of having a conda environment
 if any('CONDA' in variable for variable in os.environ):
+    print(os.environ)
     conda_list  = subprocess.run(['conda', 'list' , 'git-lfs'], stdout=subprocess.PIPE).stdout.decode('utf-8')
     if not ('git-lfs' in conda_list):
         raise ModuleNotFoundError
-
+# No conda environment
 else:
     try:
+
         import git_lfs
     except ModuleNotFoundError as error :
         raise error
--- a/ersilia/__init__.py
+++ b/ersilia/__init__.py
@@ -69,14 +69,11 @@ INSTALL_STATUS = check_install_status()["status"]
 
 __all__ = ["__version__"]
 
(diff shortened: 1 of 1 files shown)


miquelduranfrigola:
This is an interesting approach.

Do you think we could make it conda-independent? It is actually quite easy to check if git lfs is installed: http://smpte-staticwebsite-staging.s3-website-us-east-1.amazonaws.com/tutorials/git-lfs-tutorial.html

Perhaps a `git lfs` command and check output would be enough?

As for testing, it is in our roadmap to develop an `exceptions.py` submodule. This will help greatly. Also, we use `pytest` in our CI/CD actions.

Nawarrr:
I thought about this approach, but what stopped me is wither it's being ran from a python script or from a terminal directly, But I think i get it now, 
So, that's what you mean is something like this ?
ers_iss

if git-lfs isn't installed it won't be working, but I don't know wither running this command multiple times is safe, another option is running same command with -h to get command help and compare with the length.

miquelduranfrigola:
OK, this is what I meant, I think this is a good approach indeed. Would be great to place it into a function. But I like the approach in general. Many thanks!

Nawarrr:
Thanks,
I will wrap it in a function and commit the changes to the PR

Nawarrr:
I've committed the changed to the PR,

Thanks for your help and support, and sorry for not understanding what you wanted immediately, 
It's my first contribution and I'm trying my best,
Thanks :)

miquelduranfrigola (approved):
check if git lfs is installed

miquelduranfrigola:
Thanks @Nawarrr for this contribution! And no need to apologize - this is a great first contrib.

PR #187, opened 2022-04-08
[bfix] print the command output correctly to show the size in the CLI

Riyabelle25:
Currently the cmd output of 
```
url = "https://github.com/ersilia-os/{0}.git/info/lfs".format(model_id)
        cmd = "echo " + url + "| perl -ne 'print $1 if m!([^/]+/[^/]+?)(?:\.git)?$!' | xargs -I{} curl -s -k https://api.github.com/repos/'{}' | grep size"
echo(
     "The disk storage of this model in KB is"
        )
os.system(cmd) 
```
wasn't showing up in the terminal.

Debugged and fixed the above by:
```
        print (
            os.system(cmd) 
        )
```

### DIFF
--- a/ersilia/cli/commands/fetch.py
+++ b/ersilia/cli/commands/fetch.py
@@ -26,12 +26,14 @@ def fetch_cmd():
     def fetch(model, mode, dockerize):
         mdl = ModelBase(model)
         model_id = mdl.model_id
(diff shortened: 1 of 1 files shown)


Riyabelle25:
How it shows up now:
Screenshot 2022-04-08 at 3 31 15 PM

miquelduranfrigola (approved):

PR #188, opened 2022-04-08
[add] Calculates time taken for each step on fetching a model

Riyabelle25:
Solves #6 

Here, I am calculating the time taken for each step (as defined under `fetch.py`) and displaying it in the CLI.
```
    def fetch(self, model_id):
        self.model_id = model_id

        self.progress["step0_seconds"] = time.time()
        self._prepare()
        # step1_seconds = time.time() - start_seconds
        self.progress["step1_seconds"] = time.time()
        print("step 1 done")

        # step2_seconds = time.time() - step1_seconds
        self._get()
        print("step 2 done")
        self.progress["step2_seconds"] = time.time()

        # step3_seconds = time.time() - step2_seconds
        self._pack()
        print("step 3 done")
        self.progress["step3_seconds"] = time.time()

        # step4_seconds = time.time() - step3_seconds
        self._toolize()
        print("step 4 done")
        self.progress["step4_seconds"] = time.time()

        # step5_seconds = time.time() - step4_seconds
        self._content()
        print("step 5 done")
        self.progress["step5_seconds"] = time.time()

        # step6_seconds = time.time() - step5_seconds
        self._check()
        print("step 6 done")
        self.progress["step6_seconds"] = time.time()

        # step7_seconds = time.time() - step6_seconds
        self._sniff()
        print("step 7 done")
        # final_seconds = time.time() - step7_seconds
        self.progress["step7_seconds"] = time.time()

        self._success()
        print("Fetching {0} done in time: {1}s".format(model_id, abs(self.progress["step7_seconds"]-self.progress["step0_seconds"])))
        for i in reversed(range(1, len(self.progress))):
            self.progress["step{0}_seconds".format(i)] -= self.progress["step{0}_seconds".format(i-1)]
```

Screenshot 2022-04-08 at 5 06 01 PM


Riyabelle25:
Finally, as @miquelduranfrigola suggested, I am storing the `progress` results into a `<model_id>progress.JSON>` that is _currently_ being written into the project's root. 

```
        with open("{0}_progress.json".format(model_id), "w") as outfile:
            json.dump(self.progress, outfile)
        print("Progress times for each step in seconds", self.progress)
```
Of course, this JSON metadata must be stored differently.

My suggestion is to simply have a `model-metadata.json` in the root, that stores information such as the `size`, `last_used_time` as well as `progress checkpoints` of each model.

miquelduranfrigola:
Hi @Riyabelle25 wow, great work.

Before I approve the PR, may I suggest a few changes?

* Do you think we can print the seconds next to the step? For example: Step 1 done: 10 seconds
* Your `model-metadata.json` is great. Let's keep it in the project root for now. The ideal approach would be to check if commit the model-metadata.json to the corresponding github repo.

Riyabelle25:
1. Hmm, 1. is an easy addition that I will work on just now.
2. About 2, isn't pushing the `metadata.JSON `to each model's repo counter-productive, as we need the metadata (such as size, checkpoints etc) _before_ fetching the model repo itself? 
For example: The progress bar would need access to this information _during_ the fetch process itself. 

Thus my suggestion is to simply have an _all-purpose_ `model-metadata.json `in the root, that stores information such as the `size`, `last_used_time` as well as `progress checkpoints` of all of Ersilia's models

Riyabelle25:
Implemented your review @miquelduranfrigola:
Screenshot 2022-04-09 at 4 08 31 PM

Riyabelle25:
> 1. Hmm, 1. is an easy addition that I will work on just now.
> 2. About 2, isn't pushing the `metadata.JSON `to each model's repo counter-productive, as we need the metadata (such as size, checkpoints etc) _before_ fetching the model repo itself?
>    For example: The progress bar would need access to this information _during_ the fetch process itself.
> 
> Thus my suggestion is to simply have an all-purpose `model-metadata.json` in the root, that stores information such as the `size`, `last_used_time` as well as `progress checkpoints` of _all_ of Ersilia's models.
> 
> This way the necessary checkpoints of a particular model (say, `molbert`) could be parsed from the `model-metadata.json` when `ersilia fetch molbert` is called. Thus allowing us to use the same checkpoints to display progress as it downloads.
> 
> Does this make sense, forgive me if it's sounding mush 😅

This is how my suggested METADATA.json looks:
Screenshot 2022-04-09 at 4 15 50 PM

Similarly, my idea is to write here other metadata (size, last-used etc, as is being worked on in other PRs) too, under each `<model_id>`

miquelduranfrigola:
OK, @Riyabelle25 looking great overall!

I get your points about the metadata not being stored in the model github repo. But what I envisage is the following:

1) When we fetch a model, the first thing we do is *just* download the metadata file from the repo.
2) We explore the metadata file, so we have an idea of how long it is going to take to download the model, and how much space will be needed.
3) Then we start the fetch normally. Because we have time estimates, we can even add a progress bar that estimates time!

What do you think? (I must say I really appreciate your thoughts, so if you think this is not a good approach, please let me know)

Riyabelle25:
Oh I see!
I think we could explore the metdata from a model repo using the GitHub API.
Or just parse through the content using  `curl` with the model metadata file's github raw uri . (considering all the model repos are public, this is possible)

miquelduranfrigola:
Hi @Riyabelle25, exactly. GitHub API would be fine. I don't really mind downloading the file or using `curl` as you suggest. Both options are great. I would prefer GitHub API since we might have some private models temporarily.

Rufaida98:
Your @miquelduranfrigola and @Riyabelle25 solutions are amazing!!
What I meant in encoding is that using the function encode() as each string in python represented by a uni code like utf-8 or acsii so when use the function the sequence of code points converted into set of bytes so you get a sense of size for the file. But of course your solutions are much handle :)

Rufaida98:
It could be used after downloading the json file

miquelduranfrigola:
Oh, I understand now @Rufaida98 . I think we are talking about similar things. Let's do this metadata.json file and we can implement your suggestion after downloading it (before the bulky fetch commands)

Rufaida98:
@miquelduranfrigola ok 👍

Amna-28:
Hi @Riyabelle25 I see you are doing a great job here. I will be pleased to help you in it. 
Or I can work on the progress bar once you finish the time calculation.

miquelduranfrigola:
Hey. I suggest we start by writing a function that, given a model identifier (e.g. eos4e40), downloads a `metadata.json` file from its corresponding repository (https://github.com/ersilia-os/eos4e40).

A while ago I wrote a method named `download_single` in the `GitHubDownloader` class: https://github.com/ersilia-os/ersilia/blob/master/ersilia/utils/download.py. The current implementation is bad - it downloads the **full** repo and then only takes the file of interest. When I wrote the method I don't thing there was a better option. Now we can use the GitHub API, so I suggest start by updating this method.

What do you think? @Riyabelle25 this is your thread, so let us know if you want to implement it yourself or you would like to have help from @Amna-28 and/or @Rufaida98 . Up to you all!

Riyabelle25:
Apologies for the delay, my health has been bad this week😓.
Yes @miquelduranfrigola as suggested earlier, I'll use the GitHub API (or the curl, whichever suits better) to get the contents of each model's `metadata.json`

Also before the above: my code in the current linked PR has to be updated to commit and push the created `metadata.json` (during `ersilia fetch`) to that model's repository. Is this step correct @miquelduranfrigola?

Riyabelle25:
I'm recovered from the flu bug, so I'll implement this by tonight. Thanks for your advice and patience @miquelduranfrigola 🌻💛

Riyabelle25:
@miquelduranfrigola, I am using the GitHub API to commit the METADATA.json in its specified model's repo.
Like so:
```python
# in ersilia/hub/fetch/fetch.py
cmd = 'curl -X PUT -H "Accept: application/vnd.github.v3+json" -H "Authorization: token {0}" https://api.github.com/repos/ersilia-os/{1}/contents/METADATA.json -d {2}'.format(token, self.model_id, data)
run_command(cmd)
```

However, I would need the `GitHub Access Token` with the necessary write scope in order to be able to commit to each model's repo. I explored Ersilia and found the `ersilia/utils/config.py` and its `Secrets` class (that gets and stores the token as a secret in `secrets_json`), but don't understand how I could get the secret itself from there. 

Do you know how I could programmatically find each <model_id>'s repo token?

miquelduranfrigola:
Hello. I am finally coming back to this. In the last two months, I have been focused on populating the Ersilia Model Hub and I did not have time to worry about the CLI backend. I think this solution is excellent. I will not merge since I have done some refactoring in the fetch scripts. However, the logic of your solution is clear and I will incorporate it myself. Many thanks, this is a fantastic approach.

PR #189, opened 2022-04-08
Update requirements.txt

AnshuKumari197:
I think streamlit is not needed to run this hub.


miquelduranfrigola:
Hi, streamlit is not needed at this stage indeed, but will be in the near future so this is why I prefer to keep it. Good catch, though!

PR #190, opened 2022-04-08
Update README_ANSHUKUMARI.md

AnshuKumari197:
@GemmaTuron I worked on your suggestion and update as per your suggestions. Please check once and give me feedback.
I changed in requirement file as you said but @miquelduranfrigola told "Hi, streamlit is not needed at this stage indeed, but will be in the near future so this is why I prefer to keep it. Good catch, though!"

### DIFF
--- a/documentation/README_ANSHUKUMARI.md
+++ b/documentation/README_ANSHUKUMARI.md
@@ -1,3 +1,4 @@
+<div id="top" align="center">
 # Welcome to Ersillia Model Hub!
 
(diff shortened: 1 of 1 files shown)


GemmaTuron (approved):

miquelduranfrigola:
People are working on streamlit these days, so let's keep it in the requirements file. Note that it is not installed in the "slim" version of ersilia (see setup.py file).

PR #212, opened 2022-04-11
Check if Git LFS is installed

dauinh:
- Problem: When Git LFS is not installed locally, large size models are not downloaded (e.g. chemprop-antibiotic)

- Solution: Add a check function in /ersilia/__init__.py

### DIFF
--- a/ersilia/__init__.py
+++ b/ersilia/__init__.py
@@ -5,6 +5,8 @@ del _version
 
 # External imports
 import os
+import subprocess as s
+import sys
 
 # Disable GPU
 os.environ["CUDA_VISIBLE_DEVICES"] = "-1"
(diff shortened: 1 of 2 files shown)


miquelduranfrigola:
Hi I think this is great. I wonder if we could include this method in a dedicated class `GitLfsRequirement` that I've just created in: `ersilia/setup/requirements/git.py`. I would prefer to keep the `__init__.py` file as clean as possible.

What do you think?

dauinh:
Hi @miquelduranfrigola, thank you for your feedback! I also think it's a better structure! I didn't know where to put it at first...

miquelduranfrigola (approved):
Looks fantastic! Approving the PR right away

PR #222, opened 2022-04-12
Added a tutorial video on how to use the ersilia model hub.

loweyvana:


GemmaTuron (changes requested):
Hi @loweyvana 

We won't merge videos in github, can you simply keep them somewhere we can access them and link t the issue?
otherwise there will be too many documents on our github repo.

Thanks!

PR #227, opened 2022-04-13
Create Update README_busirat.md

busiratazeez:
I updated the installation guide and the license file


GemmaTuron (changes requested):
Hi @busiratazeez 

This will create a new file, rather than modifying your old one

PR #231, opened 2022-04-15
Update Readme file 

julietugo:
I created a readme file as part of the task in the Ersilia-Outreachy application.


GemmaTuron (changes requested):
The file is in the root of the repository, not under the documentation folder, can you move it?

Issue #12, opened 2022-03-23
Better model catalog display


miquelduranfrigola:
# Background
Currently, users can list the catalog of all available ersilia models with the command `ersilia catalog`. They have the option to list only models that are available on their local computer with `ersilia catalog --local`.
The current output of these commands is a table printed on the terminal using the [tabulate](https://pypi.org/project/tabulate/) python library.

# Requested features
We would like to have a better output, although it is not clear to us what would be the best option. Options we've considered are:
1. Open a browser tab and display models there.
2. Continue with tabulate (or similar) and add a multi-page view.

sayantani11:
@miquelduranfrigola Would like to give a try!

HarrietMwanza:
Hi @sayantani11 I am an Outreachy applicant and picked this first-time issue to contribute. Please do walk me over what I can help with.

Lewanja:
Hi @miquelduranfrigola , I am a first time outreachy intern and would like to contribute to this project. Please, guide me through. Where do I start?

sayantani11:
@miquelduranfrigola In order to understand the issue a bit more, I would like to know what are the problems that are being faced by the current catalog display....and what are the maintainers looking for in the future?

miquelduranfrigola:
Hi @sayantani11 I think that `tabulate` is quite simple and, perhaps, an option that displays in the terminal with a multi-page view would be better. Do you think this is easy to achieve? As the catalog grows, it will be not feasible to display everything in the command line. Suggestions welcome!

As mentioned in issue #126 I would rather discard the website option at the moment since the AirTable is sufficient for now. Also, working on a GUI at the moment (updates soon!) so let's not duplicate efforts.

anamika-yadav99:
Hi @miquelduranfrigola multi-page view can be implemented using `latex_longtable` table format which is provided by `tabulate`.  Can I implement it and raise a pr for the same?

miquelduranfrigola:
Hi @Deepthi562 the task was assigned to @sayantani11 a while ago, and I think we are on a good track with this. Do you have any suggestion that would go in the lines of what @anamika-yadav99 and @sayantani11 are discussing?

anamika-yadav99:
[...]

PR #238, opened 2022-04-16
Enhanced Model catalog display

anamika-yadav99:
Issue #12 
I have improved the model display according to discussion in comment https://github.com/ersilia-os/ersilia/issues/12#issuecomment-1096002423

I have added a function to slice the catalog table  into table of 15. Table size of 25 was huge. 
I have also added `ersilia catalog --next` and `ersilia catalog --previous` in cli which displays the next and previous table consequently.
I have done so by storing a variable called counter in an auxiliary file which is initialized to  1 everytime `ersilia catalog` is called.  Whenever `ersilia catalog --next` is called the counter is updated by 1 and displays the next table. Vise versa happens in `ersilia catalog --previous`. Counter is decreased by 1.
Implementing this was a little tricky due to all the corner cases of reaching the last table or `ersilia catalog --previous` on the 1st table. So, I have set the value of counter to 1 every time the table is on the 1st page. The user is informed about the last page  table. The counter is decreased by 2 when on the last page in order to not display the same page.  I hope all the corner cases are covered.

This is how it looks. Table size of only 5 was used here because ersilia hub has only 21 models, which made it hard to check for the corner cases:
image

image

image


Rufaida98:
@anamika-yadav99 nicely done 💯

anamika-yadav99:
There were a 2 bugs before. Which I have fixed now.
1. The counter would keep updating even after the table was ended which returned error on command `ersilia catalog --previous` and also if user did `ersilia catalog --next`
2. The initialize funtion was run even when `--next` or `--previous` command were used.

image
image
image

anamika-yadav99:
> @anamika-yadav99 nicely done 💯

Thanks @Rufaida98

Issue #241, opened 2022-04-16
Feature Request: Improve model search functionality


anamika-yadav99:
### Is your feature request related to a problem? Please describe.

yes, every once in a while there's a typo or user enters a wrong spelling. model search function in such cases return null table which could be confusing for the user. 

### Describe the solution you'd like.

I'm still exploring for a good solution.  perhaps, a simple approximate string matching algorithm which would return the closest matching model name in the ModelSearch function and also on fetch command.

anamika-yadav99:
@GemmaTuron would this be a feature you would like to have?

anamika-yadav99:
_class ErsiliaModel(ErsiliaBase):

    """Child class of ErsiliaBase which serves as the main class for Ersilia Model.
       
       This class provides methods to fetch, serve, api and more.  
       
       
       Attributes:
           model(str):
           save_to_lake(bool, optional): Option to save the calculation to Isaura Lake  
           service_class(str, optional): Service class used. Accepts only [None, conda, venv, system, docker].
           config_json(str, optional): Path to a configuration file.
           credential_json(str, optional): Path to a credentials file.
           verbose(str, optional): Set verbosity.
           fetch_if_not_avalible(bool, optional): Fectches model from the hub if not available in the local.
    """_

dauinh:
Hi @miquelduranfrigola @anamika-yadav99, can i work on this feature as well?

dauinh:
I found [this article](https://towardsdatascience.com/autocorrect-in-google-amazon-and-pinterest-and-how-to-write-your-own-one-6d23bc927c81) to be helpful. Do you think we can do after query autocorrection and implement a Trie to find the correct command?

anamika-yadav99:
Hi @dauinh Thanks for the help. The article was insightful. I'm also thinking of comparing the distance between the input model name and the model names in the hub or local and returning the closest matching model name. This is basically string matching algorithm only. Let's wait for what @miquelduranfrigola and @GemmaTuron have to say on this.

dauinh:
> Hi @dauinh Thanks for the help. The article was insightful. I'm also thinking of somewhere along the similar lines. Compare the distance between the input model name and the model names in the hub or local and then return the closest matching model name. This is basically string matching algorithm only. Let's wait for what @miquelduranfrigola and @GemmaTuron have to say on this.

Oh, i see what you mean... Thank you for clarifying!

GemmaTuron:
Hi @anamika-yadav99 

[...]

PR #262, opened 2022-05-05
ModelSearcher updated

anamika-yadav99:
issue #241   
I have added 2 methods to ModelSearcher Class to search using `--text` and `--mode` . To search using `--text` search_text method is added which also performs fuzzy match on the input to the data in catalog table. Fuzzy match is performed on model_id, description and slug.
image
image

search_mode method is kept same as previous search which searches the mode by which the model was trained. It doesn't perform any string matching.
Also, when no match is found user is asked to look through the catalog


miquelduranfrigola (approved):

miquelduranfrigola:
Hi @anamika-yadav99 this is great stuff. I have approved the PR

PR #310, opened 2022-10-11
Dependabot configured to create more PR

honeyankit:
Configure dependabot to create 10 PRs. By default it creates only 5 PRs. We can increase this even more in the future.

Increasing the PR creation config in dependabot.yml file based on the error: https://github.com/ersilia-os/ersilia/network/updates/480983059
image

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,4 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "monthly"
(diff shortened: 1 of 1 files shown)


honeyankit:
@miquelduranfrigola @GemmaTuron Also let me know if you prefer to run Dependabot version update daily, weekly or monthly. Currently I have set it up for monthly.

GemmaTuron (approved):

PR #397, opened 2022-10-20
Added template for predict model action workflow

honeyankit:
To implement an action workflow for Ersilia to execute the model and generate the prediction values.


honeyankit:
As of now, I have not added the schedule run code to run prediction model action every 10 minutes. For now, this action workflow will be invoked manually.

GemmaTuron (approved):

Issue #280, opened 2022-09-24
[Epic] Auto-test New Model Submissions


D1M1TR10S:

miquelduranfrigola:
Hello @GrantBirki and @Lehcar : I have slightly modified a previous action by @honeyankit to be potentially used to auto-test new model submissions. You can find it [here](https://github.com/ersilia-os/ersilia/actions/workflows/quick-manual-model-test.yml).

Basically what is different is that we:
* Only do 5 predictions instead of 100
* Run models on verbose mode (`-v`), so there is more log in the GitHub Actions console.

We would need to modify this action a little bit, I assume, to catch the id of the model (probably from the current github repo path, or from the README file). At the moment, the model id is specified manually.

What do you think, does this look good?

GrantBirki:
@miquelduranfrigola I think that looks great! Is this a workflow that would be helpful to have the artifact saved with log output from the logs when the test fails?

miquelduranfrigola:
Yes, I think so! From what I see, the log file is stored at `/home/runner/eos/current.log`. So if we can collect that file, that'd be awesome! Otherwise, we can still see the log in the console with the current action, so that would also be ok for now if the artifact turns out to be difficult.

Lehcar:
> Yes, I think so! From what I see, the log file is stored at `/home/runner/eos/current.log`. So if we can collect that file, that'd be awesome! Otherwise, we can still see the log in the console with the current action, so that would also be ok for now if the artifact turns out to be difficult.

I think it should be pretty straightforward. I'll see if I can add that real quick

Lehcar:
Ok, so it looks like `/home/runner/eos/current.log` is 0 bytes, but `/home/runner/eos/console.log` has data. Here are some options for the upload:

1. [Upload all `.log` files in `/home/runner/eos`](https://github.com/ersilia-os/ersilia/actions/runs/3681361624)
2. [Only upload `/home/runner/eos/console.log`](https://github.com/ersilia-os/ersilia/actions/runs/3681449466)

We could also `tee` the logs from each step: 

```
      - name: Predict output
        run: |
          source activate | tee /tmp/predict-output.log
          echo "Sample model id selected: $MODEL_ID" | tee -a /tmp/predict-output.log
          ersilia -v fetch $MODEL_ID | tee -a /tmp/predict-output.log
          ersilia -v serve $MODEL_ID | tee -a /tmp/predict-output.log
[...]

PR #501, opened 2022-12-13
Save logs as artifact

Lehcar:
Part of: https://github.com/ersilia-os/ersilia/issues/280

## Description

Uploads logs as an artifact if the action fails

## Testing
Example Runs:
Failed runs uploads artifact: 
https://github.com/ersilia-os/ersilia/actions/runs/3789468132
https://github.com/ersilia-os/ersilia/actions/runs/3789462144

Successful run doesn't upload artifact: 
https://github.com/ersilia-os/ersilia/actions/runs/3790521417

### DIFF
--- a/.github/workflows/quick-manual-model-test.yml
+++ b/.github/workflows/quick-manual-model-test.yml
@@ -37,7 +37,7 @@ jobs:
           echo "After conda init"
           conda init
           python -m pip install -e .[test]
-  
+
       - name: Predict output
         run: |
           source activate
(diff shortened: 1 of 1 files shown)


.github/workflows/quick-manual-model-test.yml
```diff
@@ -47,3 +47,9 @@ jobs:
           ersilia sample -n 5 -f input.csv
           ersilia -v api -i input.csv
           ersilia close
+
+      - name: Upload log output
```
GrantBirki, line 51:
```suggestion
      - name: Upload log output
        if: ${{ failure() }}
```

Since we only want to upload an artifact with error logs, should we gate this step to only run when something fails?

GrantBirki (approved):

PR #504, opened 2022-12-14
Adding a dockerfile and an easy-install script.

ttokarczuk:
Added an easyinstall script in dockerfiles/easyinstallimage
This script is meant for the automation of running ersilia in docker.
It also creates an alias to ersilia in the host system.
How to use:
1. Run script dockerersilia.sh
2 The script will:
- Create a directory in ~/ersiliafiles of the host system. The directory is mounted in the Docker image in /root/eos directory.
- Try to fetch Ersilia docker image and start it.
- Add an alias to the shell rc file: alias ersilia="docker exec -it ersilia /bin/bash" - so everytime you type ersilia in the host system shell it will enter Ersilia docekr container.

### DIFF
--- a/dockerfiles/easyinstallimage/Dockerfile
+++ b/dockerfiles/easyinstallimage/Dockerfile
@@ -1,0 +1,21 @@
+FROM ubuntu:20.04 
+SHELL ["/bin/bash", "-c"]
+RUN apt-get update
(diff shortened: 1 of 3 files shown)


miquelduranfrigola (approved):
Reviewed. Looks great.

Issue #540, opened 2023-01-13
🦠 Model Request: testing


GrantBirki:
### Model Name

test name

### Model Description

test description goes here

### Slug

test-model

### Tags

test,something,blah

### License

MIT

GrantBirki:
I am just testing things out here

GemmaTuron:
This is now fixed @GrantBirki !

PR #541, opened 2023-01-13
Metadata Workflow Addition

GrantBirki:
# Metadata Workflow Addition 🔢 

This pull request adds to the new model submission flow to populate the newly created repo's `metadata.json` file with values parsed from the new model issue template

This PR also adds a new workflow called `json syntax check` which will validate the syntax to make sure it is valid for **all** JSON file types in this repo for extra safety

## About 💡 

Here is a brief explanation about how this works:

1. The new model submission workflow already parses the issue template into JSON so we save this as a GitHub Actions output to use in subsequent steps
2. A later step in the workflow creates the new repository for the model
3. **new** - we then run a workflow step that parses both the `metadata.json` file that came over from the template repository (now found in the 'new' repo) and the issue template (json format)
4. **new** - this workflow step executes a script that mashes these together to populate the `metadata.json` file and commits it to the new repo
5. **new** - a new workflow step reads the updated `metadata.json` file and executes the Airtable Python script by @miquelduranfrigola to ship data about the repo to Airtable
6. The workflow continues normally with an issue comment posted back to the user, etc

## Examples 📷 

See examples below

### Before

image

### After

image

> Parsed from an example issue https://github.com/ersilia-os/ersilia/issues/540

---

## Merge Note 🚀 

Since our workflow only runs when changes land on the `main` branch, we will need to merge this PR to fully test. I have tested all that I can locally and it _should_ work 🤞 

I will be doing my testing from this issue: https://github.com/ersilia-os/ersilia/issues/540

---

related: https://github.com/ersilia-os/ersilia/issues/280, https://github.com/ersilia-os/github-partners/issues/3


miquelduranfrigola (approved):
Great approach, thanks!

PR #573, opened 2023-01-20
Pushing changes to download.py and config.py to enable downloads from…

ttokarczuk:
… S3 bucket.
- Adding s3 download functions
- Removing _clone_with_gh and _clone_with_pygit2 to allow only Git downloads.
- Updating config.py to point to an S3 bucket.


miquelduranfrigola (approved):
This looks fantastic, thanks so much, @ttokarczuk !

Issue #626, opened 2023-03-08
🐛 Bug: Comment on New Model request form


GemmaTuron:
### Describe the bug.

When we approve a new model request using the /approve command, the  repository is updated and the record is created in the airtable, but the nice reply to the contributor in the issue does not appear.
It fails at the last step, which uses a Git Action created by @GrantBirki .
See this workflow: https://github.com/ersilia-os/ersilia/actions/runs/4364036742/jobs/7630862860

The only thing I am aware that changed is that I moved @GrantBirki from owner to external contributor since the period of work with GitHub finished --- might this be the source? 

### Describe the steps to reproduce the behavior

Check the workflow run https://github.com/ersilia-os/ersilia/actions/runs/4364036742/jobs/7630862860

### Expected behavior.

A comment to be created in the model request issue

### Operating environment

Ubuntu 20.04 LTS

GrantBirki:
Taking a look 👀

GrantBirki:
It looks like the `GrantBirki/comment` Action is failing and its pointing to this section in the readme -> https://github.com/GrantBirki/comment#error-resource-not-accessible-by-integration-

GrantBirki:
While looking at the workflow, I have a few suggestions and will open a pull request

GrantBirki:
I'm also looking into https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

GemmaTuron:
Thansk for checking this so fast!
It's solved now https://github.com/ersilia-os/ersilia/issues/580

PR #627, opened 2023-03-08
Approve Updates / Fixes

GrantBirki:
Pin's exact release tags and sha's for Actions stability and security. It also updates the `permissions:` block at the top of the repo to try and fix the bug where the final step fails to post a comment on the issue that triggered the Action workflow

related: https://github.com/ersilia-os/ersilia/issues/626


GrantBirki:
@GemmaTuron We can merge this at your earliest convenience an then test the `/approve` workflow again

PR #722, opened 2023-06-22
Corrected the  is_installed typo

samuelmaina:
renamed is_installed  to _is_installed to solve the installation bug.
Was able to fetch install ersilia successfully.


miquelduranfrigola:
Thanks @samuelmaina - this indeed solves the issue. I fixed it without seeing your PR, so I won't merge it to avoid redundancy. But your fix was perfect. Thanks.

Issue #825, opened 2023-10-03
🐛 Bug: Error when running 'ersilia -v api run -i <input>' command


PromiseFru:
### Describe the bug.

When I run the following command:

```
ersilia -v api run -i "CCCC"
```

I encounter the following error:

```
Traceback (most recent call last):

File "/home/Ersilia/ersilia/ersilia/io/readers/file.py", line 321, in read_input_columns
if len(h) == 1:
TypeError: object of type 'NoneType' has no len()
```

### Describe the steps to reproduce the behavior

1. Run the command: `ersilia -v api run -i "CCCC"`
2. Observe the error.

### Expected behavior.

I expect that running the `ersilia -v api calculate -i "CCCC"` command should calculate the molecular weight of the input molecule ("CCCC") and display the result in the CLI as follows:

```
{
    "input": {
        "key": "IJDNQMDRQITEOD-UHFFFAOYSA-N",
        "input": "CCCC",
        "text": "CCCC"
    },
    "output": {
        "mw": 58.123999999999995
    }
}
```

### Operating environment

Ubuntu 20.04.6 LTS (Focal Fossa)

### Additional context

I found that modifying line 321 in `/ersilia/io/readers/file.py` from:

```python
if len(h) == 1:
```

to:

```python
if h is not None and len(h) == 1:
```

resolves the issue.

DhanshreeA:
Hi @PromiseFru I suspect you ran into this issue while running the `eos3b5e` model. Just a quick tip, while creating issues here, it's useful for the maintainers if you mention the model id/ model name.

Additionally, you should not need to modify anything within the Ersilia CLI codebase. For resolving this issue, please refer to the wonderful discussion here: https://github.com/ersilia-os/ersilia/issues/820. If you are still unable to get the model to work, please use your Contribution Period issue to log any errors you encountered.

Issue #820, opened 2023-10-02
✍️ Contribution period: Leila Yesufu


leilayesufu:
### Week 1 - Get to know the community

- [X] Join the communication channels
- [X] Open a GitHub issue (this one!)
- [x] Install the Ersilia Model Hub and test the simplest model
- [x] Write a motivation statement to work at Ersilia
- [x] Submit your first contribution to the Outreachy site

### Week 2 - Install and run an ML model

- [x] Select a model from the suggested list
- [x] Install the model in your system
- [x] Run predictions for the EML
- [x] Compare results with the Ersilia Model Hub implementation!
- [x] Install and run Docker!

### Week 3 - Propose new models

- [x] Suggest a new model and document it (1)
- [x] Suggest a new model and document it (2)
- [x] Suggest a new model and document it (3)

### Week 4 - Prepare your final application

- [x] Submit the final application in the Outreachy website

leilayesufu:
<img width="906" alt="Screenshot 2023-10-02 203111" src="https://github.com/ersilia-os/ersilia/assets/78659709/97d75a64-88b8-41f5-ac52-7a82d35b3973">
Today, 2nd of October 2023.
I got accepted into the outreachy contribution stage and chose to contribute to Ersilia Open Source Initiative project.
I'm very excited to learn and grow during this contribution stage.

**Task 1: Join the communication channels** 
I joined the slack workspace and introduced myself in the general Channel.

**Task 2:  Open a GitHub issue (this one!)**
Opened the issue. Done :)

leilayesufu:
**Task 3:  Install the Ersilia Model Hub and test the simplest model**
**October 2nd 2023**

Following the instructions [here](https://ersilia.gitbook.io/ersilia-book/ersilia-model-hub/installation).

I installed the Ersilia Model hub from the Command line Interface.
I'm using a windows device so i had to use the windows subsytem for Linux with an Ubuntu terminal.

There are prerequisites to be installed before downloading the Model hub. This can be seen as follows.
1.   **Install the gcc compiler on Ubuntu terminal**
        This can be down with by running the following
         ` sudo apt-get update`
          `sudo apt install build-essential`
           <img width="388" alt="gcc" src="https://github.com/ersilia-os/ersilia/assets/78659709/55b70dc5-96ae-4f2c-87d0-7739c17bddff">
The gcc compiler has been successfully installed.


2.  **Install  Python and Conda**
     I installed the Miniconda using the following command.
[...]

PR #827, opened 2023-10-03
updated read_input_columns method in /ersilia/io/readers/file.py

PromiseFru:
Modified line 321 in /ersilia/io/readers/file.py from:

```
if len(h) == 1:
```
to:

```
if h is not None and len(h) == 1:
```

resolves issue #825 .


DhanshreeA:
Hi @PromiseFru generally you should not need to change anything within the Ersilia codebase. Please refer to the discussion [here](https://github.com/ersilia-os/ersilia/issues/820) to get Ersilia up and running.

Issue #919, opened 2023-12-19
🐛 Bug: Poetry installation in the PyPi workflow


miquelduranfrigola:
### Describe the bug.

In trying to update Ersilia to PyPi using GitHub Actions, we get the following error: https://github.com/ersilia-os/ersilia/actions/runs/7259408946/job/19776488322

### Operating environment

GitHub Actions

DhanshreeA:
The bug is introduced because the poetry installer script has been moved to a different location than the URL used in the workflow file. Ref: https://github.com/python-poetry/poetry#installation

As the tool has matured over time, the authors of Poetry recommend using a manual install with pip pinned to a specific version as part of CI recommendations. Ref: https://python-poetry.org/docs/#ci-recommendations
Creating a PR with the same.

miquelduranfrigola:
@DhanshreeA now we are getting this error. Should be an easy fix (twine not found): https://github.com/ersilia-os/ersilia/actions/runs/7260110067/job/19778596014

GemmaTuron:
This is fixed here: https://github.com/ersilia-os/ersilia/pull/920

GemmaTuron:
Sorry went too quickly
Shouldn't twine be installed with poetry? see the current error: https://github.com/ersilia-os/ersilia/actions/runs/7260110067/job/19778596014

GemmaTuron:
This is now solved

DhanshreeA:
@GemmaTuron could we actually test this? I don't know how to test this other than to create a tag. :sweat_smile:

GemmaTuron:
we actually want to make a new release because we have made many updates, so we could test it this way @miquelduranfrigola do you think?

miquelduranfrigola:
Yes, let me do a tag

PR #921, opened 2023-12-19
Use Poetry to publish to PyPI directly

DhanshreeA:
Forego using Twine to publish. Use poetry directly to build and publish (which also kind of justifies the use of poetry in the first place)


DhanshreeA:
Hopefully closes #919 :grimacing:

GemmaTuron (commented):
seems good!

PR #927, opened 2023-12-21
Add a new tracking module to enable the monitoring of runs via Splunk

AC-Dap:
This PR is the result of the work by Harvard T4SG to help Ersilia begin monitoring their runs via Splunk. Namely, we accomplish the following:

- A new module located at `ersilia/core/tracking.py` to handle all the logic of calculating and uploading statistics
- Captures both output metrics (summary statistics, mismatched types, etc.) and run metrics (runtime, memory usage, etc.)
- By default, tracking is disabled; it is enabled with a single CLI flag, `ersilia run -t`
- Automatic uploading of metrics through AWS S3 into Splunk
- Some initial skeleton code to also upload outputs to CDD Vault, which was suggested to enable simple visualization of molecular data. Although this code is unfinished, the remaining TODOs should not be too difficult.

We also have some initial work on persistent tracking across an entire session, which includes parsing the log file for any warnings or errors. This is enabled by a flag when serving the model—`ersilia serve -t`—but we currently do not upload the resulting files to AWS and Splunk. Whether this is also a desirable data source is left to Ersilia to decide.


miquelduranfrigola:
It looks great. Thanks so much!

Issue #930, opened 2023-12-26
🐛 Bug: Model repo created and not eliminated if action fails


GemmaTuron:
### Describe the bug.

When approving a new model, the workflow might fail if, for example, the contributor changed a field in the form. When it gets an approve command, the repo will be created  and just sit there unused, as the metadata won't be committed, the model di will not be updated on airtable nor the user will receive a response in the issue.
We should add a check that if the action fails, it looks for the created repository and deletes it.
Not prioritary, but nice to have

### Operating environment

Ubuntu 20.04 LTS

miquelduranfrigola:
I see. Do you have an example of this? A link would suffice

miquelduranfrigola:
Interesting. I agree, this would be nice to have. Let me quickly tag @GrantBirki and see if he has an idea on how to address this.

GemmaTuron:
This is now sloved with the user and Ersilia's admins getting a warning about the failure of the workflow and pointing to the repo to be eliminated, thanks!

PR #938, opened 2024-01-02
add a comment for the user on failure

GrantBirki:
This pull request leaves a comment for the user with relevant links if the `/approve` workflow fails. It didn't feel too safe to be deleting repos directly from an automated workflow run so I chose not to run those operations here. Rather, this just lets the user know a failure occurred with direct links to the orphaned repo and the workflow logs.

related: https://github.com/ersilia-os/ersilia/issues/930

### DIFF
--- a/.github/templates/approve-workflow-failed.md
+++ b/.github/templates/approve-workflow-failed.md
@@ -1,0 +1,7 @@
+### Workflow Failure ❌
+
+@{{ issue_creator }} the `/approve` workflow has failed. View the logs here for more information:
+
+🔗 [Workflow logs]({{ workflow_url }})
+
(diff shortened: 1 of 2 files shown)


miquelduranfrigola:
Hello @GrantBirki ! Thanks as always for a very quick response. This looks like a very good solution. @GemmaTuron, what do you think? Do we have an easy way to test it with our current models in progress?

@GrantBirki - I am merging the PR. We will let you know in case something arises but, overall, this looks great. Thanks so much.

GemmaTuron:
That is great thanks @GrantBirki !
I'll test it right away

miquelduranfrigola:
Awesome @GemmaTuron - please let us know how it went.

GemmaTuron:
I answered in the open issue, @GrantBirki fix works like a charm!

PR #961, opened 2024-02-09
Remove newline characters in Airtable Output

carcablop:
The CSV file generated previously had more rows than expected due to the presence of newline characters. To address this issue, the newline characters were removed using the re.sub() function.


miquelduranfrigola:
Hello @carcablop I have merged the PR.

@GemmaTuron - I think that, looking forward, we want to remove newline characters in any field in the model catalog. We could control for this in the Information class of Ersilia. Let me know if you think this is worth solving and I will create a spike issue accordingly

PR #1112, opened 2024-04-23
Configure model docker builds to be rebuilt with older dockerfiles

DhanshreeA:
This is a very monkey patched fix, and there are several things that can be improved here but in the interest of time, this is going on the backburner right now.


miquelduranfrigola:
Thanks.
Ideally, we should not use two images from Ersilia. For now it is OK.

PR #1129, opened 2024-05-23
hotfix: pin requests to a version less than 2.31.x

DhanshreeA:
Requests which is currently set to ^2.31.0 installs 2.32.x which breaks docker_py as tracked [here](https://github.com/docker/docker-py/issues/3256).


DhanshreeA:
PS this is only a temporary fix, ideally we would not want to restrict requests like that, and update the docker version when they make a new release (hopefully soon)

Issue #1090, opened 2024-03-25
[🐅 Epic]: Splunk Integration into Ersilia


DhanshreeA:
### Summary
Ersilia is host to a number of models that only keeps growing. With the advancements in Generative AI in drug discovery, we are now incorporating compute intensive models into the model hub. Ersilia's primary stakeholders are scientists and researchers working with limited resources and therefore it is imperative that we collect performance metrics related to a model's run after it is incorporated in Ersilia. This work has multiple aspects:

- the actual collection of a model run metrics
- persisting these metrics in a remote collection
- visualization of these metrics in a dashboard (powered by Splunk Enterprise) 
- monitoring and visualizing these metrics at scale per model (this will be enabled in conjunction through the work done by GDI at https://github.com/ersilia-os/model-inference-pipeline)

We want to test and integrate the `ersilia run --track` flag more closely within our workflows and use of the Splunk powered Ersilia Model TA dashboard as a way to closely monitor models within the Ersilia Model Hub. 

### Objective(s)

The project is divided into code refactoring tasks and then various phases of Splunk integration, and finally tasks related to making the dashboard accessible over the web. These tasks can be executed independent of each other unless specified otherwise.

**Prerequisites**

- [x] Test the track flag locally to see if it produces the desired results on your machine

**Code Refactoring**:
- [x] Remove Pandas as a dependency by using the natively in built csv module instead (#928 and #1054)
- [x] Resolve duplication between `RunLogger` in `ersilia/core/session.py` and `RunTracker` in `ersilia/core/tracking.py`. For example, both RunTracker and RunLogger subsample result output when the output dimensions are too many to reduce storage overhead, and to not clutter the monitoring dashboard. https://github.com/ersilia-os/ersilia/pull/1132
- [x] Remove multiple, redundant, tracking flags and and keep a single boolean flag with a default False value such that by default we do not track runs for now. https://github.com/ersilia-os/ersilia/pull/1160
- [x] Rework code calculating memory usage - output and Docker container https://github.com/ersilia-os/ersilia/issues/1162
- [x] Bug: The tracking flag does not currently work when a CSV file is not specified for storing the model predictions. https://github.com/ersilia-os/ersilia/issues/1164

**Phased Integration**:
- [ ] Phase 0.1: Integrate tracking command within Ersilia model test pipeline prior to refactoring and benchmark this (#925)
- [ ] Phase 0.2: Stress testing a model and computing performance and resource usage with different input scenarios and systems
[...]

PR #1132, opened 2024-05-31
Resolve duplication between RunLogger in ersilia/core/session.py and RunTracker in ersilia/core/tracking.py

Malikbadmus:
**Description**

Resolve duplication between RunLogger in ersilia/core/session.py and RunTracker in ersilia/core/tracking.py. For example, both RunTracker and RunLogger subsample result output when the output dimensions are too many to reduce storage overhead and to not clutter the monitoring dashboard.

**Changes Made**

- Runlogger Class merged to the RunTracker classes in tracking.py,  creating a new unified class that inherits from both base classes and includes the initialization logic for both.
- Created a single __init__ method that handles the initialization requirement for tracking and logging.
- Dropped the `sample_df` method, and modified modified most of the functions in the RunTracker class. 
- Created a `track_run` function with clearer method names and simplified parameters, excluding input and result data frames, meta, and metadata from the JSON output.
- Changed the session file path to be under ersilia_runs.
- Modified `model.py` to reflect the changes.

**Status**

- The tracking module works with the changes made.
- The size of the JSON object has been significantly reduced (more than 10 times), and potential issues with large DataFrames have been avoided.
- The session file is now rerouted to the same directory used by the rest of the logging and tracking files.
- The memory usage during the model run was also reduced.

Related to https://github.com/ersilia-os/ersilia/issues/1090

### DIFF
--- a/ersilia/core/tracking.py
+++ b/ersilia/core/tracking.py
@@ -401,8 +401,6 @@ class RunTracker(ErsiliaBase):
             column_stats["max"] = max(values)
             column_stats["std"] = statistics.stdev(values) if len(values) > 1 else 0
 
-            stats[column] = column_stats
-
         return stats
         
         
--- a/ersilia/core/tracking.py
+++ b/ersilia/core/tracking.py
@@ -1,19 +1,23 @@
-from datetime import datetime
+import os
 import json
+import sys
+import shutil
 import csv
+import logging
+import tracemalloc
 import statistics
 from collections import defaultdict
-import sys
-import tracemalloc
+from datetime import datetime
+from .base import ErsiliaBase
+from ..io.output_logger import TabularResultLogger
+from ..default import EOS, ERSILIA_RUNS_FOLDER 
 import tempfile
-import logging
 import boto3
 from botocore.exceptions import ClientError, NoCredentialsError
-import os
 import re
 import requests
 
-PERSISTENT_FILE_PATH = os.path.abspath("current_session.txt")
+
 # Temporary path to log files until log files are fixed
 TEMP_FILE_LOGS = os.path.abspath("")
 
@@ -81,52 +85,72 @@ def log_files_metrics(file):
                 if misc_error_flag:
                     errors[error_name] += 1
 
-        write_persistent_file(f"Error count: {error_count}")
+        write_persistent_file(f"Error count: {error_count}", model_id)
         if len(errors) > 0:
-            write_persistent_file(f"Breakdown by error types:")
+            write_persistent_file(f"Breakdown by error types:", model_id)
             for error in errors:
-                write_persistent_file(f"{error}: {errors[error]}")
-        write_persistent_file(f"Warning count: {warning_count}")
+                write_persistent_file(f"{error}: {errors[error]}", model_id)
+        write_persistent_file(f"Warning count: {warning_count}", model_id)
     except (IsADirectoryError, FileNotFoundError):
         logging.warning("Unable to calculate metrics for log file: log file not found")
-        
+
+
+
+def persistent_file(model_id):
+
+    """
+    Create the persistent path file 
+    :param model_id: Te currently running model
+    """
+
+    persistent_file_dir = os.path.join(EOS, ERSILIA_RUNS_FOLDER, "session", model_id)
+    os.makedirs(persistent_file_dir, exist_ok=True)  
+    file_name = os.path.join(persistent_file_dir, "current_session.txt")
+    
+    return file_name
+
+    
 
 def open_persistent_file(model_id):
     """
     Opens a new persistent file, specifically for a run of model_id
     :param model_id: The currently running model
     """
-    with open(PERSISTENT_FILE_PATH, "w") as f:
+  
+    file_name = persistent_file(model_id)
+    with open(file_name, "w") as f:
         f.write("Session started for model: {0}\n".format(model_id))
         
         
-def write_persistent_file(contents):
+def write_persistent_file(contents, model_id):
+
     """
     Writes contents to the current persistent file. Only writes if the file actually exists.
     :param contents: The contents to write to the file.
     """
-
-    # Only write to file if it already exists (we're meant to be tracking this run)
-    if os.path.isfile(PERSISTENT_FILE_PATH):
-        with open(PERSISTENT_FILE_PATH, "a") as f:
+    
+    file_name = persistent_file(model_id)
+    if file_name and os.path.isfile(file_name):
+        with open(file_name, "a") as f:
             f.write(f"{contents}\n")
 
 
-def close_persistent_file():
+def close_persistent_file(model_id):
     """
     Closes the persistent file, renaming it to a unique name.
     """
-
-    # Make sure the file actually exists before we try renaming
-    if os.path.isfile(PERSISTENT_FILE_PATH):
+        
+    file_name = persistent_file(model_id)
+    if os.path.isfile(file_name):
         log_files_metrics(TEMP_FILE_LOGS)
 
         new_file_path = os.path.join(
-            os.path.dirname(PERSISTENT_FILE_PATH),
+            os.path.dirname(file_name),
             datetime.now().strftime("%Y-%m-%d%_H-%M-%S.txt"),
         )
-        os.rename(PERSISTENT_FILE_PATH, new_file_path)
+        os.rename(file_name, new_file_path)
         
+
 
 def upload_to_s3(json_dict, bucket="ersilia-tracking", object_name=None):
     """Upload a file to an S3 bucket
@@ -228,6 +252,7 @@ def upload_to_cddvault(output_df, api_key):
         return False
 
 
+
 def read_csv(file_path):
     """
     Reads a CSV file and returns the data as a list of dictionaries.
@@ -239,6 +264,8 @@ def read_csv(file_path):
         reader = csv.DictReader(file)
         data = [row for row in reader]
     return data
+    
+    
 
 def get_nan_counts(data_list):
         """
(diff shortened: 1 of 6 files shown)


ersilia/core/tracking.py
```diff
@@ -225,166 +242,170 @@ def upload_to_cddvault(output_df, api_key):
...
+	"""
         self.time_start = datetime.now()
         tracemalloc.start()
         self.memory_usage_start = tracemalloc.get_traced_memory()[0]
-
-    def sample_df(self, df, num_rows, num_cols):
-        """
-        Returns a sample of the dataframe, with the specified number of rows and columns.
+    
+    def calculate_stats(self, result):
```
DhanshreeA, line 282: I don't think you should be modifying this method in this PR - it's unrelated and already being worked on by @dzumii

ersilia/core/tracking.py
```diff
@@ -225,166 +242,170 @@ def upload_to_cddvault(output_df, api_key):
...
+            stats[column] = {
+                "mean": df[column].mean(),
+                "mode": df[column].mode().iloc[0] if not df[column].mode().empty else None,
+                "min": df[column].min(),
+                "max": df[column].max(),
+                "std": df[column].std(),
+            }
+        return stats    
+
+    def get_file_sizes(self, df):
```
DhanshreeA, line 308: Same as above - don't modify this method in this PR. Unrelated. Being worked on by @dzumii

ersilia/core/tracking.py
```diff
@@ -225,166 +242,170 @@ def upload_to_cddvault(output_df, api_key):
...
+	Calculates the size of the input and output dataframes, as well as the average size of each row.
+	:df: Pandas dataframe containing the data
+	:return: dictionary containing the input size, output size, average input size, and average output size
+	"""
+    
+        size = df.memory_usage(deep=True).sum() / 1024
+        avg_row_size = size / len(df)
+        return size, avg_row_size
+
+    def validate_data(self, df, metadata):
```
DhanshreeA, line 319: Nope. Don't modify in this PR. Unrelated.

ersilia/core/tracking.py
```diff
@@ -225,166 +242,170 @@ def upload_to_cddvault(output_df, api_key):
...
+            json.dump(meta, f)
+
+    def log_logs(self):
+        output_dir = os.path.join(self.logs_folder, self.model_id)
+        os.makedirs(output_dir, exist_ok=True)
+        file_name = os.path.join(output_dir, f"{self.model_id}.log")
+        session_file = os.path.join(EOS, "session.json")
+        shutil.copyfile(session_file, file_name)
+
+    def track_run(self, input_path, result_path, meta):
```
DhanshreeA, line 372: Nope. Don't modify this either. We'll get to it in a different issue.

DhanshreeA (changes requested):

ersilia/core/tracking.py
```diff
@@ -89,22 +99,29 @@ def log_files_metrics(file):
 
 
 def open_persistent_file(model_id):
+
     """
     Opens a new persistent file, specifically for a run of model_id
     :param model_id: The currently running model
     """
+    global PERSISTENT_FILE_PATH
```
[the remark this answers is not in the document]
Malikbadmus: I was looking for a way to pass values to maintain and update the file path across multiple (open_persistent_file, write_persistent_file, and close_persistent_file) functions.

So when open_persistent_file is called, the path to the persistent file is updated for the specific model_id,  and calls to `write_persistent_file(contents)` will use this updated value to write the contents to the correct persistent file.

ersilia/core/tracking.py
```diff
@@ -1,16 +1,26 @@
...
 import tempfile
-import logging
 import boto3
 from botocore.exceptions import ClientError, NoCredentialsError
-import os
 import re
 import requests
 
-PERSISTENT_FILE_PATH = os.path.abspath("current_session.txt")
+ERSILIA_RUNS_FOLDER = "ersilia_runs"
```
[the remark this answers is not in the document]
Malikbadmus: Okay, I will do that

Malikbadmus:
> Hey @Malikbadmus, good work so far and thank you for your efforts and the detailed explanations here! I have a few questions.
> 
> You mention:
> 
> > The size of the JSON object has been significantly reduced (more than 10 times), and potential issues with large DataFrames have been avoided.
> 
> 1. Which JSON object? The one containing all the tracking information?
> 2. What potential issues with large dataframes?
> 
> > The memory usage during the model run was also reduced.
> > I understand that this might be happening because we have a single class now instead of two classes, however:
> 
> 1. This is not the memory usage we are interested in anyway. We are interested in the memory used by the actual model however this code is currently getting the memory used by the ersilia CLI python process running the model. However, I agree the code could benefit from optimization, which brings me to my second point about this...
> 2. How much was the reduction? Do you have estimates/results on that? If yes, could you please share those here?
> 
> > The session file is now rerouted to the same directory used by the rest of the logging and tracking files.
> > This is useful, thank you!

@DhanshreeA , Many thanks for the review.

1. Yes, I meant the result of the tracking run, the variable  `json_object`.
2. Since the `json_object` contains the full Input and result data frame, which can be large depending on the input file we use, excluding it should reduce memory consumption and improve performance.
3. The peak memory used before the refactor was 4.01 MB, after 1.135. 

[2024-05-2818-33-15.txt](https://github.com/user-attachments/files/15545713/2024-05-2818-33-15.txt)

DhanshreeA:
> 2\. Since the `json_object` contains the full Input and result data frame, which can be large depending on the input file we use, excluding it should reduce memory consumption and improve performance.

Thanks @Malikbadmus, you're right this is absolutely not required, your suggestion is helpful.

Malikbadmus:
with @Inyrkz assistance , a better alternative to the Global variable has been implemented.

Malikbadmus:
@DhanshreeA , I have successfully rebased the pull request to work with the latest changes made to the master branch and resolve the conflict that emerged.

ersilia/core/tracking.py
```diff
@@ -81,53 +85,73 @@ def log_files_metrics(file):
...
-        write_persistent_file(f"Warning count: {warning_count}")
+                write_persistent_file(f"{error}: {errors[error]}", model_id)
+        write_persistent_file(f"Warning count: {warning_count}", model_id)
     except (IsADirectoryError, FileNotFoundError):
         logging.warning("Unable to calculate metrics for log file: log file not found")
-        
+
+
+
+def persistent_file(model_id):
```
DhanshreeA, line 99: Rename to "create_persistent_file"

ersilia/core/tracking.py
```diff
@@ -81,53 +85,73 @@ def log_files_metrics(file):
...
+
+    """
+    Create the persistent path file 
+    :param model_id: Te currently running model
```
DhanshreeA, line 103: Typo: "The" currently running model

DhanshreeA (changes requested):

DhanshreeA:
LGTM @Malikbadmus, just two more very small changes and we can merge this

ersilia/core/tracking.py
```diff
@@ -81,53 +85,73 @@ def log_files_metrics(file):
...
+
+    persistent_file_dir = os.path.join(EOS, ERSILIA_RUNS_FOLDER, "session", model_id)
+    os.makedirs(persistent_file_dir, exist_ok=True)  
+    file_name = os.path.join(persistent_file_dir, "current_session.txt")
+    
+    return file_name
+
+    
 
 def open_persistent_file(model_id):
```
DhanshreeA, line 114: You can get rid of this function entirely and move the print statement here to `create_persistent_file`

DhanshreeA (changes requested):

Malikbadmus:
> LGTM @Malikbadmus, just few more very small changes and we can merge this

@DhanshreeA , I have modified the PR to reflect these changes.

ersilia/core/tracking.py
```diff
@@ -81,53 +85,65 @@ def log_files_metrics(file):
...
     """
     Writes contents to the current persistent file. Only writes if the file actually exists.
     :param contents: The contents to write to the file.
     """
-
-    # Only write to file if it already exists (we're meant to be tracking this run)
-    if os.path.isfile(PERSISTENT_FILE_PATH):
-        with open(PERSISTENT_FILE_PATH, "a") as f:
+    
+    file_name = create_persistent_file(model_id)
```
DhanshreeA, line 124: I'm sorry I don't understand why we need to create the file before writing to it or closing it every time? This write function is supposed to write in append mode only when the file already exists. A better approach would be to have a different `check_file_exists` method if that is what we are trying to ensure here.

ersilia/core/tracking.py
```diff
@@ -81,53 +85,65 @@ def log_files_metrics(file):
...
-def close_persistent_file():
+def close_persistent_file(model_id):
     """
     Closes the persistent file, renaming it to a unique name.
     """
-
-    # Make sure the file actually exists before we try renaming
-    if os.path.isfile(PERSISTENT_FILE_PATH):
+        
+    file_name = create_persistent_file(model_id)
```
DhanshreeA, line 135: Same as above... it does not make much sense to create a file first only to then close it. My apologies if I missed this part earlier.

DhanshreeA (changes requested):

ersilia/core/tracking.py
DhanshreeA: Same as above, I would recommend using a `check_file_exists` type of method to check the file exists before closing it instead of creating so we can close it.

Malikbadmus:
The new error handling I added to the function `close_persistent_file` is being thrown up, this is because  if a model is run without the tracking flag, a persistent file is not going to be created, therefore the function `check_file_exist` will return false.

This might also be related to this issue #1155

ersilia/core/tracking.py
```diff
@@ -110,38 +110,52 @@ def create_persistent_file(model_id):
...
-    file_name = create_persistent_file(model_id)
-    if file_name and os.path.isfile(file_name):
+    if check_file_exists(model_id):
+        file_name = os.path.join(EOS, ERSILIA_RUNS_FOLDER, "session", model_id, "current_session.txt")
         with open(file_name, "a") as f:
             f.write(f"{contents}\n")
 
-
+    else:
+        raise FileNotFoundError(f"The persistent file for model {model_id} does not exist. Cannot write contents.")
```
Inyrkz, line 138: Great job adding the `check_file_exists()` function @Malikbadmus.

Here's my thought. Wouldn't it be better to create the file and write the content to it, if the file doesn't exist in the `else` block instead of raising an error?

Malikbadmus: Many thanks @Inyrkz .

My thought was an error will be more appropriate, since it indicated something might have gone wrong or a process has been bypass, leading to a rerun of the processes. 

Write_persistent_file() is being used by many methods, adding create_file_path function in the else block might lead to a session file being created with insufficient information, or what do you think?

DhanshreeA: I think I'm in the favor of raising an error rather than silently ignoring it so I am with @Malikbadmus on this one.

DhanshreeA:
LGTM @Malikbadmus merging this!

PR #1150, opened 2024-06-05
Bump requests and docker-py

### LINKED ISSUE #1146: 🐈 Task: Bump requests and docker-py versions in ersilia
### Summary

From a dependabot alert:

When making requests through a Requests Session, if the first request is made with verify=False to disable cert verification, all subsequent requests to the same origin will continue to ignore cert verification regardless of changes to the value of verify. This behavior will continue for the lifecycle of the connection in the connection pool.
Remediation

Any of these options can be used to remediate the current issue, we highly recommend upgrading as the preferred mitigation.

    Upgrade to requests>=2.32.0.
    For requests<2.32.0, avoid setting verify=False for the first request to a host while using a Requests Session.
    For requests<2.32.0, call close() on Session objects to clear existing connections if verify=False is used.

Related Links

    https://github.com/psf/requests/pull/6655

### Objective(s)

- [ ] Bump requests to 2.32.0
- [ ] Bump docker_py to 7.1.0

### Documentation

https://github.com/psf/requests/issues/6707
https://github.com/psf/requests/pull/6710

DhanshreeA:
Closes #1146


miquelduranfrigola:
What is the status of this PR?

Issue #1156, opened 2024-06-09
🐛 Bug: Fetching models on MacBook (M1) results in 404 error due to looking for linux/arm64


kartikey-vyas:
### Describe the bug.

Running `ersilia -v fetch eos2r5a` (same outcome for any model ID) results in:

```
404 Client Error for http+docker://localhost/v1.45/images/ersiliaos/eos2r5a:latest/json: Not Found ("No such image: ersiliaos/eos2r5a:latest")
```

Conditions:
- macbook pro, with M1 Pro chip
- fresh install of ersilia in a clean virtual env, 
- fully updated docker desktop, v4.30.0

I think I know whats wrong. 

By default, `fetch` will eventually call `ModelPuller` which tries to pull from DockerHub via a command like:
```
docker pull ersiliaos/eos2r5a:latest
```

The output from this command when it can't find the image is this log:
```
no matching manifest for linux/arm64/v8 in the manifest list entries
```

The regex command which is used to search this log to determine whether to raise `DockerConventionalPullError` is 
```
re.search(r"no match.*platform.*manifest", pull_log):
```
This command won't find anything since there is no matching string `platform` in the error log, so it will just assume it was successful. It should be an easy fix and I'm happy to do it, but I thought I'd raise it here first to make sure this is undesirable behaviour.

### Describe the steps to reproduce the behavior

1. run `ersilia -v fetch eos2r5a` on an apple silicon device

### Operating environment

macOS Sonoma 14.4.1

DhanshreeA:
Hey @kartikey-vyas thank you for opening this issue. It's interesting to me that on a my Mac (M3 Pro, Docker desktop v4.30.0 and v4.31.0) I see the error string:
```
no match for platform in manifest: not found
```
However I agree, removing platform would generalize this exception better. I'll go ahead and merge your PR.

PR #1157, opened 2024-06-09
fix: relax regex search of docker logs

kartikey-vyas:
Addresses #1156 

Docker output log from failed pull on apple silicon is:

```
no matching manifest for linux/arm64/v8 in the manifest list entries
```

current regex is:
```
r"no match.*platform.*manifest"
```

there's no 'platform' in this message so we won't match; we should remove the match on 'platform'


DhanshreeA:
LGTM!

PR #1158, opened 2024-06-10
Remove multiple and redundant track flags

dzumii:
**Description**

Remove multiple, redundant, tracking flags and keep a single boolean flag with a default false value, such that by default we do not track runs for now.

**Changes to be made**

Change the `track_serve` flag to `track` such that it tracks the session without the need to use track_run.
Remove the `track_run` flag 
Compare the output with the output of the former state.

**Status**
Successfully changed the multiple track flags into one (`track`)
track flag is only needed at the point of running the model

This is still a work in progress

Related to #1090

### DIFF
--- a/ersilia/cli/commands/close.py
+++ b/ersilia/cli/commands/close.py
@@ -4,7 +4,7 @@ from . import ersilia_cli
 from .. import echo
 from ... import ErsiliaModel
 from ...core.session import Session
-from ...core.tracking import close_persistent_file
+from ...core.tracking import check_file_exists, close_persistent_file
 
 
 def close_cmd():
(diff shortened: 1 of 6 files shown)


dzumii:
Hi @DhanshreeA 
I have tried several ways to achieve what we want but right now, what I have done is to remove the `track` flag from `run.py` but the `track` value will be `true` by default. This way, once the `track` flag is used while serving the model, the run sessions will be tracked by default, and if the `track` flag is not used, the persistent_file won't be opened and the metric file won't be written. I know this is not efficient.
The other way I thought of is to save the `track` value to the `config.json` file and import the track value from there into the `run.py `file.
This way, I won't have to declare a default value for `track` in the `run.py` file. 
I could use some suggestions here

DhanshreeA:
Hey @dzumii could you please resolve the conflicts in your branch? We could pair over it if you'd like.

dzumii:
> Hey @dzumii could you please resolve the conflicts in your branch? We could pair over it if you'd like.

Alright

PR #1160, opened 2024-06-14
Remove multiple and redundant track flags

dzumii:
**Description**

Remove multiple, redundant tracking flags and keep a single boolean flag with a default false value, such that by default we do not track runs for now.

**Changes to be made**

Change the `track_serve` flag to `track` and ensure it tracks the session without the need to use `track_run` again.
Remove the `track_run` flag
Compare the output with the output of the former state.

**Status**
Successfully changed the multiple track flags into one (`track`)
track flag has been moved to serve

**To do**
Confirm that the metrics of several run sessions of the same model are appended to the same file

Related to #1158 #1090

### DIFF
--- a/ersilia/core/model.py
+++ b/ersilia/core/model.py
@@ -49,7 +49,7 @@ class ErsiliaModel(ErsiliaBase):
         fetch_if_not_available=True,
         preferred_port=None,
         log_runs=True,
-        track_status=False,
+        track_runs=False,
     ):
         ErsiliaBase.__init__(
             self, config_json=config_json, credentials_json=credentials_json
@@ -81,7 +81,7 @@ class ErsiliaModel(ErsiliaBase):
             "hosted",
         ], "Wrong service class"
         self.service_class = service_class
-        self.track_status=track_status
+        self.track_runs = track_runs
         mdl = ModelBase(model)
         self._is_valid = mdl.is_valid()
         assert (
@@ -133,7 +133,7 @@ class ErsiliaModel(ErsiliaBase):
         else:
             self._run_logger = None
 
-        if track_status:
+        if track_runs:
             self._run_tracker = self._run_logger
         else:
             self._run_tracker = None
@@ -422,10 +422,9 @@ class ErsiliaModel(ErsiliaBase):
 
     def serve(self):
         self.close()
-        self.session.open(model_id=self.model_id,track_status=self.track_status)
+        self.session.open(model_id=self.model_id, track_runs=self.track_runs)
         self.autoservice.serve()
         self.session.register_service_class(self.autoservice._service_class)
-        self.session.register_tracking_status(self.track_status)
         self.url = self.autoservice.service.url
         self.pid = self.autoservice.service.pid
         self.scl = self.autoservice._service_class
@@ -548,4 +547,3 @@ class ErsiliaModel(ErsiliaBase):
     @property
     def _model_info(self):
         return self.info()
-
--- a/ersilia/core/model.py
+++ b/ersilia/core/model.py
@@ -49,7 +49,7 @@ class ErsiliaModel(ErsiliaBase):
         fetch_if_not_available=True,
         preferred_port=None,
         log_runs=True,
-        track_runs=False,
+        track_status=False,
     ):
         ErsiliaBase.__init__(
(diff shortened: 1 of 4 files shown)


dzumii:
Hello @DhanshreeA 
I've been able to remove redundant track flags. Now, users only have to use the track flag at the point of serving the model alone and don't need to use it for every model run. Also, I confirmed the metrics for each run are saved to the same session [file](https://github.com/user-attachments/files/15836875/current_session.txt). For example, I ran the model six times with different inputs, and the metrics for the six runs are in the txt file.

ersilia/core/session.py
```diff
@@ -42,12 +42,26 @@ def register_service_class(self, service_class):
...
+        with open(self.session_file, "w") as f:
+            json.dump(data, f, indent=4)
+
+    def open(self, model_id,track_status):
         self.logger.debug("Opening session {0}".format(self.session_file))
         session = {
             "model_id": model_id,
             "timestamp": str(time.time()),
             "identifier": str(uuid.uuid4()),
+            "track_status":track_status
```
DhanshreeA, line 64: Add space after key. In fact, just format the file with black.

ersilia/core/session.py
```diff
@@ -42,12 +42,26 @@ def register_service_class(self, service_class):
...
 
-    def open(self, model_id):
+    def tracking_status(self):
+        data = self.get()
+        if data is None:
+            return None
+        else:
+            return data["track_status"]
+
+    def register_tracking_status(self, track_status):
```
DhanshreeA, line 52: Do we really need this method if we are serializing `track_status` when we first write to the session JSON? I don't think we need this method.

ersilia/core/model.py
```diff
@@ -421,9 +422,10 @@ def update_model_usage_time(self, model_id):
 
     def serve(self):
         self.close()
-        self.session.open(model_id=self.model_id)
+        self.session.open(model_id=self.model_id,track_status=self.track_status)
         self.autoservice.serve()
         self.session.register_service_class(self.autoservice._service_class)
+        self.session.register_tracking_status(self.track_status)
```
DhanshreeA, line 428: Same, this is unnecessary since we are already writing `track_status` to session JSON.

ersilia/core/model.py
```diff
@@ -49,7 +49,7 @@ def __init__(
         fetch_if_not_available=True,
         preferred_port=None,
         log_runs=True,
-        track_runs=False,
+        track_status=False,
```
DhanshreeA, line 52: I think we can either keep the variable name as `track` or `track_runs` as it previously was.

ersilia/core/model.py
```diff
@@ -421,9 +422,10 @@ def update_model_usage_time(self, model_id):
 
     def serve(self):
         self.close()
-        self.session.open(model_id=self.model_id)
+        self.session.open(model_id=self.model_id,track_status=self.track_status)
```
DhanshreeA, line 425: Let's keep it as `track`, or `track_runs`.

ersilia/core/session.py
```diff
@@ -42,12 +42,26 @@ def register_service_class(self, service_class):
...
+        data = self.get()
+        data["track_status"] = track_status
+        with open(self.session_file, "w") as f:
+            json.dump(data, f, indent=4)
+
+    def open(self, model_id,track_status):
```
DhanshreeA, line 58: `track`, or `track_runs`.

DhanshreeA (changes requested):

DhanshreeA:
@dzumii LGTM, requesting few minor changes.

ersilia/core/session.py
dzumii: Done

ersilia/core/session.py
dzumii: You are right, i have removed it

ersilia/core/model.py
dzumii: Right, i have made changes

ersilia/core/model.py
dzumii: Done

ersilia/core/session.py
dzumii: Done

DhanshreeA (approved):

PR #1131, opened 2024-05-30
Code Refactor: Remove Pandas as a dependency by using the natively in built csv module instead

dzumii:
**Description**
Code refactoring to remove Pandas from the tracking functionality and implement the functionality using built-in Python libraries.

**Changes Made**
- defined `read_csv` function to read CSV files using the native CSV library.
- Replaced some other pandas data manipulation.
- Adjusted the `stats` function to calculate statistics using the statistics library.
- Adjusted the `get_file_size` function

**Status**
- The tracking functionality works with the changes. 
- Time taken is now a bit shorter.
- However there are still some inconsistencies in the outcome in comparison with the outcome generated using the tracking functionality with pandas, particularly relating to the file_size and stats functions.

**To Do**
- Refine the `check_types` function.
- Refine the `get_file_size` function.

This is still a Work In Progress
Comments and suggestions would be really appreciated

Related to #1090 

Closes #1054

PR #1170, opened 2024-06-20
Track without output bug

### LINKED ISSUE #1164: 🐛 Bug: Tracking functionality does not work when an output is not specified
### Describe the bug.

When you serve a model without the track flag and without specifying the output, the output is printed in the CLI. But when you serve the model with the track flag without specifying the output, an error is returned.

1. Fetch a model
2. Serve it with the track flag
3. Run the model without specifying the output file

Related to #1090

### Operating environment

Codespace

dzumii:
**Description**
If track flag is used without specifying an output, an error is returned. This is because tracking requires a result parameter, which is the output of a run session. The error is thrown because the output being passed from run into the parameter is a generator object and not a csv file.

**Changes Made**
Wrote the generator file into a csv file, which is now read into the track function.

**Status**
Tracing now works without specifying output

**To do**
Make sure track flag works with all output formats (csv, tsv, json and h5) supported by ersilia 

Related to #1131 
Closes #1164

### DIFF
--- a/ersilia/core/tracking.py
+++ b/ersilia/core/tracking.py
@@ -22,7 +22,6 @@ from ..io.output_logger import TabularResultLogger
 from botocore.exceptions import ClientError, NoCredentialsError
 
 
-
 def log_files_metrics(file_log, model_id):
     """
     This function will log the number of errors and warnings in the log files.
@@ -61,7 +60,9 @@ def log_files_metrics(file_log, model_id):
                     # encountering new logs
                     # make sure error flags are closed
                     if ersilia_error_flag:
-                        errors["Unknown Ersilia exception class"] = errors.get("Unknown Ersilia exception class", 0) + 1
+                        errors["Unknown Ersilia exception class"] = (
+                            errors.get("Unknown Ersilia exception class", 0) + 1
+                        )
                         ersilia_error_flag = False
                     if misc_error_flag:
                         errors[error_name] = errors.get(error_name, 0) + 1
@@ -86,15 +87,15 @@ def log_files_metrics(file_log, model_id):
                     errors["Unknown Ersilia exception class"] += 1
                 if misc_error_flag:
                     errors[error_name] += 1
-        
+
         json_dict = {}
         json_dict["Error count"] = error_count
-        
+
         if len(errors) > 0:
             json_dict["Breakdown by error types"] = {}
             for error in errors:
-                json_dict["Breakdown by error types"][error] = errors[error]      
-        json_dict["Warning count"] = warning_count      
+                json_dict["Breakdown by error types"][error] = errors[error]
+        json_dict["Warning count"] = warning_count
         json_object = json.dumps(json_dict, indent=4)
         write_persistent_file(json_object, model_id)
     except (IsADirectoryError, FileNotFoundError):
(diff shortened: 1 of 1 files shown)


DhanshreeA:
@dzumii could you please resolve the conflicts on this branch? Thank you!

dzumii:
> @dzumii could you please resolve the conflicts on this branch? Thank you!

Done

dzumii:
@DhanshreeA 
The tracking functionality now works without an output specified and with the csv, tsv and json output formats

ersilia/core/tracking.py
```diff
@@ -275,53 +272,57 @@ def upload_to_cddvault(output_df, api_key):
...
 def read_csv(file_path):
     """
     Reads a CSV file and returns the data as a list of dictionaries.
 
     :param file_path: Path to the CSV file.
     :return: A list of dictionaries containing the CSV data.
     """
-    with open(file_path, mode='r') as file:
+
+    with open("output.csv", "w", newline="") as csvfile:
```
DhanshreeA, line 283: Maybe we can write this file outside either directly in the track function or have a new function called something like "write_result_csv_when_not_a_file". Right now this function is misleading for two reasons:
1. It's writing inside when it's supposed to only read csvs
2. Every time the function is invoked, it will create a file called 'output.csv' even when an input file (or input) is passed.

DhanshreeA (changes requested):

ersilia/core/tracking.py
dzumii: I have made changes to this, the csv file is written inside the track function if result is a generator object

ersilia/core/tracking.py
```diff
@@ -536,7 +537,27 @@ def track(self, input, result, meta):
         self.docker_client = SimpleDocker()
         json_dict = {}
         input_data = read_csv(input)
-        result_data = read_csv(result)
+        # Create a temporary file to store the result if it is a generator
```
DhanshreeA, line 540: Reword to "create a temporary file to store the result if..."

ersilia/core/tracking.py
```diff
@@ -536,7 +537,27 @@ def track(self, input, result, meta):
...
+        if isinstance(result, types.GeneratorType):
+
+            # Ensure EOS/tmp directory exists
+            tmp_dir = os.path.join(EOS, "tmp")
+            if not os.path.exists(tmp_dir):
+                os.makedirs(tmp_dir, exist_ok=True)
+
+            # Create a temporary file to store the generator output
```
DhanshreeA, line 548: Reword similarly as above.

DhanshreeA:
@dzumii LGTM! Minor documentation changes but rest is good!

ersilia/core/tracking.py
dzumii: I am not sure i got you well. What you want me to reword it to looks the same to me
"create a temporary file to store the result if it is a generator"

ersilia/core/tracking.py
dzumii: Do you mean I should reword to "create a temporary file to store the result if it is a generator" same as above?

ersilia/core/tracking.py
DhanshreeA: Oh my bad, I think I got confused a bit when I read the word output at L548. I think this is okay, let me mark this conversation as resolved and merge this.

DhanshreeA (approved):

Issue #1072, opened 2024-03-13
🐛 Bug: Ersilia Catalog when no models are available


GemmaTuron:
### Is your feature request related to a problem? Please describe.

When you try to run `ersilia catalog --local` but no models are available locally, instead of telling you this it outputs an error:

```
Traceback (most recent call last):
  File "/home/gturon/anaconda3/envs/ersilia/bin/ersilia", line 33, in <module>
    sys.exit(load_entry_point('ersilia', 'console_scripts', 'ersilia')())
  File "/home/gturon/anaconda3/envs/ersilia/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/home/gturon/anaconda3/envs/ersilia/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/home/gturon/anaconda3/envs/ersilia/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/gturon/anaconda3/envs/ersilia/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/gturon/anaconda3/envs/ersilia/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/home/gturon/github/ersilia-os/ersilia/ersilia/cli/commands/__init__.py", line 22, in wrapper
    return func(*args, **kwargs)
  File "/home/gturon/github/ersilia-os/ersilia/ersilia/cli/commands/catalog.py", line 47, in catalog
    catalog = mc.local().as_json()
AttributeError: 'NoneType' object has no attribute 'as_json'
```

### Describe the solution you'd like.

We could add a better info - Not prioritary at all though

Adhivp:
can you assign me this issue @DhanshreeA ?, I would be happy to learn and work on it

GemmaTuron:
Hello @Adhivp 

If you are working on the Outreachy contribution period please follow the guidelines and the tasks set out on the specific issue, we are a small community and cannot provide support outside those.
Thanks for your understanding

Adhivp:
It's ok , hope to see support in the future, Have a nice day

GemmaTuron:
[...]

PR #1175, opened 2024-06-26
Ersilia Catalog Bug when no models are available

dzumii:
**Description**
When you try to run ersilia catalog --local but no models are available locally, instead of telling you there is no local model available, it outputs an error:

**Changes Made**
-updated the `local` method to return an empty `catalogue_table` if no models are found.
-updated the `catalogue` command to print a message when the `catalogue_table` has no data

**Status**
"No local model is available. Please fetch a model by running 'ersilia fetch ...'" is returned when you run  `ersilia catalog --local` and no local model is available 

Related to #1072

### DIFF
--- a/ersilia/cli/commands/catalog.py
+++ b/ersilia/cli/commands/catalog.py
@@ -1,6 +1,7 @@
 import click
 
 from . import ersilia_cli
+from .. import echo
 from ...hub.content.catalog import ModelCatalog
 from ...hub.content.search import ModelSearcher
 
@@ -32,7 +33,8 @@ def catalog_cmd():
     def catalog(local=False, file_name=None, browser=False, more=False):
         if local is True and browser is True:
             click.echo(
-                "You cannot show the local model catalog in the browser", fg="red"
+                "You cannot show the local model catalog in the browser",
+                fg="red",
             )
         if more:
             only_identifier = False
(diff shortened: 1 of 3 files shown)


ersilia/cli/commands/catalog.py
```diff
@@ -43,15 +45,25 @@ def catalog(local=False, file_name=None, browser=False, more=False):
             mc.airtable()
             return
         if local:
+            catalog_table = mc.local()
```
DhanshreeA, line 48: I like your use of storing `mc.local` and `mc.hub` into a `catalog_table` variable. Could we maybe rewrite this whole bit as:

```Python

catalog_table = mc.local() if local else mc.hub()

if file_name is None:
    catalog = catalog_table.as_json()
else:
    catalog_table.write(file_name)
    catalog = None
```

DhanshreeA (changes requested):

ersilia/cli/commands/catalog.py
```diff
@@ -1,6 +1,7 @@
 import click
 
 from . import ersilia_cli
+from .. import echo
```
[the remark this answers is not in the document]
dzumii: Oh! I was only trying something out,forgot to remove it

DhanshreeA (approved):

DhanshreeA:
LGTM!

ersilia/cli/commands/catalog.py
dzumii: Alright

PR #1183, opened 2024-07-01
Refactor tracking to support a single SMILES string input

Malikbadmus:
**Description**

Making tracking work when input is just a single SMILE string.

**Changes Made**

- Modified and moved `read_csv` method to `utils/csv`, and created a dictionary method to process generator output before storing to a CSV file.
- A minor fix in `model.py` to enable tracking if `standard run` was enabled.

**Status**
Tracking works with different input/output formats.

Related to #1170


DhanshreeA:
LGTM!

PR #1195, opened 2024-07-10
Modifying the local Catalog functionality

Malikbadmus:

**Description**

Utilizing `information.json` instead of the  AirTable database to populate the local model card.

**Changes to be made**

- Modified the dir path in the two (2) `LocalCard` class method.
- Made use of lru_cache to cache the results.

**Status**

`ersilia catalog --local --more` command now takes between 0.5s-0.8s, representing an almost 97% improvement in performance.

Related to #1072


ersilia/default.py
```diff
@@ -30,6 +30,7 @@
 LOGGING_FILE = "console.log"
 CURRENT_LOGGING_FILE = "current.log"
 CARD_FILE = "card.json"
+LOCAL_CARD_FILE = "information.json"
```
DhanshreeA, line 33: We already have a file called INFORMATION_FILE, do we need to create another one?

ersilia/hub/content/card.py
```diff
@@ -45,7 +46,7 @@
 except:
     Hdf5Explorer = None
 
-from ...default import CARD_FILE, METADATA_JSON_FILE, SERVICE_CLASS_FILE
+from ...default import EOS, LOCAL_CARD_FILE, METADATA_JSON_FILE, SERVICE_CLASS_FILE
```
DhanshreeA, line 49: I don't understand where 'LOCAL_CARD_FILE' is being written to for your code to be able to read it. If it is the same information file as i mentioned above, in that case, you can reuse that same identifier instead of creating a new one.

DhanshreeA (changes requested):

Malikbadmus:
@DhanshreeA , you are right, there's no need to create a new file, when we have an existing file that serves the same function.

I've modified the PR to reflect this.

ersilia/hub/content/card.py
```diff
@@ -704,30 +706,33 @@ def get(self, model_id):
 class LocalCard(ErsiliaBase):
     def __init__(self, config_json):
         ErsiliaBase.__init__(self, config_json=config_json)
-
+    
+    @lru_cache(maxsize=32)
     def get(self, model_id):
-        model_path = self._model_path(model_id)
-        card_path = os.path.join(model_path, CARD_FILE)
+        model_path = os.path.join(EOS,  "dest", model_id)
```
DhanshreeA, line 712: Why can't we use 'self._model_path?' here? Is there a particular reason for rewriting that to explicitly resolving the model path like this?

ersilia/hub/content/card.py
```diff
@@ -704,30 +706,33 @@ def get(self, model_id):
...
         """
         This method returns information about how the model was fetched by reading 
         the service class file located in the model's bundle directory. If the service 
         class file does not exist, it returns None.
         """
-        service_class_path = os.path.join(
-                self._get_bundle_location(model_id), SERVICE_CLASS_FILE
-            )
+        model_path = os.path.join(EOS,  "dest", model_id)
+        service_class_path = os.path.join(model_path, INFORMATION_FILE)
```
DhanshreeA, line 730: I think it's better to call it 'information_file_path' because calling it 'service_class_path' is misleading since it is not the path to the SERVICE_CLASS_FILE, but to the INFORMTION_FILE.

DhanshreeA (changes requested):

DhanshreeA:
Looks pretty good @Malikbadmus , thanks for implementing all the suggestions. Merging this now.

DhanshreeA (approved):

PR #1205, opened 2024-07-22
Adding model source descriptor to the information file of model fetched from Docker

### LINKED ISSUE #1196: 🐈 Task: Reflect correct values in information.json
### Summary

Presently, when a model is fetched from DockerHub, the `information.json` file, which we rely upon for retrieving general characteristics about a fetched model, is [copied](https://github.com/ersilia-os/ersilia/blob/e6edabeaad9f1dff80062c5699dfa690e98ff2fb/ersilia/hub/fetch/lazy_fetchers/dockerhub.py#L45) from the container storage to the host storage - however this is incorrect since it is only a representation of the model environment inside the container and not the one on the host machine. 

Eg: If I fetch `eos3b5e` from DockerHub, as per the information.json made available on my machine, the model's size is <1MB, and its service class is `conda`. However, the model size is ~750 MB (the model image size), and the model's service class is `pulled_docker`. 

The `information.json` file is currently being used in the following ways:

1. [In ErsiliaModel.info()](https://github.com/ersilia-os/ersilia/blob/e6edabeaad9f1dff80062c5699dfa690e98ff2fb/ersilia/core/model.py#L545)
2. [In LocalCard._load_data()](https://github.com/ersilia-os/ersilia/blob/e6edabeaad9f1dff80062c5699dfa690e98ff2fb/ersilia/hub/content/card.py#L715)
3. [ModelInformer.__init__()](https://github.com/ersilia-os/ersilia/blob/e6edabeaad9f1dff80062c5699dfa690e98ff2fb/ersilia/hub/fetch/actions/inform.py#L16)
4. [ModelHostedFetcher.get_information()](https://github.com/ersilia-os/ersilia/blob/e6edabeaad9f1dff80062c5699dfa690e98ff2fb/ersilia/hub/fetch/lazy_fetchers/hosted.py#L81) and ModelHostedFetcher.get_metadata()
5. [PulledDockerImageService._get_memory()](https://github.com/ersilia-os/ersilia/blob/e6edabeaad9f1dff80062c5699dfa690e98ff2fb/ersilia/serve/services.py#L525)
6. [StandardCsvRunApi.get_input_type()](https://github.com/ersilia-os/ersilia/blob/e6edabeaad9f1dff80062c5699dfa690e98ff2fb/ersilia/serve/standard_api.py#L111) and StandardCsvRunApi.is_output_type.standardizable()
7. Ersilia test

### Objective(s)

TODO

### Documentation

_No response_

Malikbadmus:

**Description**

Adding model source value to the information file copied from container storage.

**Changes to be made**

Including model source descriptor in modify_information method.

**Status**

Information file now displays where a model was fetched from.

**To do**

Update the catalog local command.

Closes #1196

### DIFF
--- a/ersilia/setup/requirements/bentoml.py
+++ b/ersilia/setup/requirements/bentoml.py
@@ -4,6 +4,7 @@ import tempfile
 import os
 import shutil
 
+from ...utils.logging import make_temp_dir
 
 class BentoMLRequirement(object):
     def __init__(self):


Malikbadmus:
Writing the source file value after fetching won't work since the value is being used during fetching operation, my workaround was to make the source file a temporary file that is deleted after being read from.

DhanshreeA:
> Writing the source file value after fetching won't work since the value is being used during fetching operation, my workaround was to make the source file a temporary file that is deleted after being read from.

@Malikbadmus I'm a bit confused, please help me understand what do you mean here?

Malikbadmus:
> > Writing the source file value after fetching won't work since the value is being used during fetching operation, my workaround was to make the source file a temporary file that is deleted after being read from.
> 
> 
> 
> @Malikbadmus I'm a bit confused, please help me understand what do you mean here?

Okay so, the model_source_file creation was moved to after fetching(when the model has been registered in the file system) i.e self._model_path is no longer null.

But we are using this same file to get the value for model source in content/information.py during fetching when the file does not yet exist.

DhanshreeA:
> > > Writing the source file value after fetching won't work since the value is being used during fetching operation, my workaround was to make the source file a temporary file that is deleted after being read from.
> > 
> > 
> > @Malikbadmus I'm a bit confused, please help me understand what do you mean here?
> 
> Okay, so what's happening is that model_source_file is being created after fetching as can be seen [here](https://github.com/Malikbadmus/ersilia/blob/552b053b3b66f74fdb383749d30db5d284c87760/ersilia/hub/fetch/fetch.py#L244) i.e when self._model_path is no longer null.
> 
> But we need this same file to be able to get the value for model source during fetching operation as seen [here](https://github.com/Malikbadmus/ersilia/blob/552b053b3b66f74fdb383749d30db5d284c87760/ersilia/hub/content/information.py#L48) but this would not be possible as this specific file does not exist in that path as at that time.

@Malikbadmus I really don't think we need this, because even though the source file does not exist when the ModelInformer code runs, after the model is fetched, it would be available on the disk, and ready to be consumed by the catalog command.

Malikbadmus:
> > > > Writing the source file value after fetching won't work since the value is being used during fetching operation, my workaround was to make the source file a temporary file that is deleted after being read from.
> 
> > > 
> 
> > > 
> 
> > > @Malikbadmus I'm a bit confused, please help me understand what do you mean here?
> 
> > 
> 
> > Okay, so what's happening is that model_source_file is being created after fetching as can be seen [here](https://github.com/Malikbadmus/ersilia/blob/552b053b3b66f74fdb383749d30db5d284c87760/ersilia/hub/fetch/fetch.py#L244) i.e when self._model_path is no longer null.
> 
> > 
> 
> > But we need this same file to be able to get the value for model source during fetching operation as seen [here](https://github.com/Malikbadmus/ersilia/blob/552b053b3b66f74fdb383749d30db5d284c87760/ersilia/hub/content/information.py#L48) but this would not be possible as this specific file does not exist in that path as at that time.
> 
> 
> 
> @Malikbadmus I really don't think we need this, because even though the source file does not exist when the ModelInformer code runs, after the model is fetched, it would be available on the disk, and ready to be consumed by the catalog command. 

Oh..okay, so we have the catalog command query MODEL_SOURCE_FILE for the model source value, then we don't need to add model source value to the model information file.

I'll refactor to reflect this.

PR #1212, opened 2024-07-23
Make MODEL_SIZE_FILE always go in the model dest directory

DhanshreeA:


miquelduranfrigola:
I think this makes sense but I did not have the chance to review it. Do you need my feedback?

miquelduranfrigola (approved):

miquelduranfrigola:
@DhanshreeA feel free to merge

Malikbadmus:
@DhanshreeA and @miquelduranfrigola, before merging, I wanted to explain why I used the current path. The data for the model size and model source is being written before the model has been registered in the file system.

As a result, `self._model_path(self.model_id)` and `self._get_bundle_location(model_id)` are null when the I/O operations are being performed.

This will cause the error No such file or directory: '/home/eos/dest/eos3b5e/size.json' when attempting to write to size.json

DhanshreeA:
> @DhanshreeA and @miquelduranfrigola, before merging, I wanted to explain why I used the current path. The data for the model size and model source is being written before the model has been registered in the file system.
> 
> As a result, `self._model_path(self.model_id)` and `self._get_bundle_location(model_id)` are null when the I/O operations are being performed.
> 
> So for model fetched from docker,this will cause the error `No such file or directory: /home/eos/dest/eos3b5e/size.json` when attempting to write to size.json

Hi @Malikbadmus - indeed, I ran into those exact issues while trying to fix this, and my work around is:
1. Writing the source file in eos/dest/model-id/ AFTER the model is fetched (which also registers the model on the file system). 
2. I am reading the model size from `ModelPuller` in place instead of first saving that value on the file system and then reading it from that file.

Issue #1222, opened 2024-08-05
🐈 Task: Refactor models with run.sh script


kurysauce:
### Summary

The updated [test script ]( https://github.com/kurysauce/ersilia_test/blob/master/ersilia/publish/test.py )reveals models that do not have an existing run.sh script. @GemmaTuron and @miquelduranfrigola agree that the solution is to refactor models to include `run.sh`. Running the update test script on a model will yield the error `"Check halted. Either run.sh file does not exist, or model was not fetched via --from_github or --from_s3."`, if a model does not already have a `run.sh` file.

### Objective(s)

Refactor models to include a `run.sh` file. Models `eos4e40` and `eos3b5e ` do not have a `run.sh` file

miquelduranfrigola:
Thank you @kurysauce !

For now, I am tagging here @HarmonySosa, @LauraGomezjurado and @DhanshreeA and we'll see who has capacity to jump into this issue.

Two comments:
1. `eos3b5e` should be easy to refactor and perhaps @HarmonySosa can help here, since she is dealing with it in the context of  `ersilia-pack`?
2. `eos4e40` should not be too difficult to refactor either, and perhaps we can gain inspiration from a related model [eos3804](https://github.com/ersilia-os/eos3804/tree/main/model/framework)?

DhanshreeA:
Also adding @Malikbadmus and @dzumii here.

dzumii:
Ok, checking this out

kurysauce:
Currently working on `eos4e40` with @miquelduranfrigola

kurysauce:
Log on local bash run for `eos4e40` shows that outputs are being captured.
```
(eos4e40) @kurysauce ➜ /workspaces/eos4e40/model/framework (main) $ bash run.sh . /workspaces/eos4e40/model/framework/input.csv /workspaces/eos4e40/model/framework/pred.csv
Reading file /workspaces/eos4e40/model/framework/input.csv
1 molecules read!
Running model!
content after reading the file: ['CC(C)(C)c1cc(O)c(cc1O)C(C)(C)C']
[...]

Issue #1221, opened 2024-08-05
🐛 Bug: Inconsistent Output Error with model eos4tcc


kurysauce:
### Describe the bug.

When model is tested using the command `ersilia -v test eos4tcc`, the `check_consistent_output` method runs the model twice with the same input and compares the outputs. An error is raised stating that both outputs have a greater than 5% difference, which is unexpected.  

Log output:
```
**20:35:56 | DEBUG    | Outputs that raise error: 0.43146425, 0.53953654. Difference: 0.10807228999999996
20:35:56 | DEBUG    | Percentage difference for outcome: 22.25997982967655**


**{'input': {'key': 'XNSAINXGIQZQOO-SRVKXCTJSA-N', 'input': 'NC(=O)[C@@H]1CCCN1C(=O)[C@H](Cc1c[nH]cn1)NC(=O)[C@@H]1CCC(=O)N1', 'text': 'NC(=O)[C@@H]1CCCN1C(=O)[C@H](Cc1c[nH]cn1)NC(=O)[C@@H]1CCC(=O)N1'}, 'output': {'outcome': [0.43146425, 0.22777487, 0.017527986]}}
{'input': {'key': 'XNSAINXGIQZQOO-SRVKXCTJSA-N', 'input': 'NC(=O)[C@@H]1CCCN1C(=O)[C@H](Cc1c[nH]cn1)NC(=O)[C@@H]1CCC(=O)N1', 'text': 'NC(=O)[C@@H]1CCCN1C(=O)[C@H](Cc1c[nH]cn1)NC(=O)[C@@H]1CCC(=O)N1'}, 'output': {'outcome': [0.53953654, 0.22860871, 0.019828169]}}**


{'input': {'key': 'IBAQFPQHRJAVAV-ULAWRXDQSA-N', 'input': 'OCCN1C[C@H](O)[C@@H](O)[C@H](O)[C@H]1CO', 'text': 'OCCN1C[C@H](O)[C@@H](O)[C@H](O)[C@H]1CO'}, 'output': {'outcome': [0.44287112, 0.22686765, 0.019868646]}}
[...]

PR #1224, opened 2024-08-08
Refactor Ersilia test command

kurysauce:

**Description**

Refactor the Ersilia test command to run in 100% completion for correctly incorporated models. Currently stops prematurely at the `check_example_input` test method.

**Changes to be made**

Ensure correct `return` values and flags are appropriately passed to methods in `Model Tester` class. Debug other methods as test command runs closer to completion. 

EDIT: Completed

**Status**

1. Added correct returns and flags to all methods
2. Fixed path parsing in `run_bash` method
3. Corrected original model size calculation implementation by implementing `get_directories_sizes()` method and associated helper functions.
4. Updated `check_consistent_output` method by calculating Spearman's Correlation & Mean Relative Absolute Value thresholds, replacing 5% difference threshold due to false positive failed test cases
5. Refactored `read_csv` method to handle differences in column parsing with Ersilia output and Bash output
6. Specified output file generation + time stamp at completion of a successful test run on model

**To do**

Must test updated command on multiple models to ensure robustness. Ensure diversified outputs are tested. Test command may reveal hidden flaws from previous model incorporations (non-existent run.sh files, inconsistent outputs comparisons, execution errors in run.sh files that reveal problems in model's `main.py` script, etc...), open issues for specific models as needed. Seeking help from community members to test models!

Models must be fetched using `--from_github` or `--from_s3` flag, or else test script will fail.

Related to #1203 , #1222, #1221


ersilia/publish/test.py
```diff
@@ -335,6 +327,7 @@ def check_information(self, output):
 
     @throw_ersilia_exception
     def check_single_input(self, output):
+        print(f"check_single_input called with output: {output}")
```
DhanshreeA, line 330: Remove or replace with debug logs

ersilia/publish/test.py
```diff
@@ -355,19 +348,19 @@ def check_single_input(self, output):
 
     @throw_ersilia_exception
     def check_example_input(self, output):
+        print(f"check_example_input called with output: {output}")
```
DhanshreeA, line 351: Remove or replace with debug logs

ersilia/publish/test.py
```diff
@@ -379,24 +372,32 @@ def check_example_input(self, output):
...
     outputs equals the number of inputs.  
     """
-
+    
+    
+    
     @throw_ersilia_exception
     def check_consistent_output(self):
-        # self.logger.debug("Confirming model produces consistent output...")
+        def compute_mrae(values1, values2):
```
DhanshreeA, line 380: I'm a bit confused, are you trying to calculate Mean absolute error or Relative absolute error? Could you point me to the formula that you are using? Also why not use a standard implementation within numpy/scipy?

DhanshreeA (changes requested):

.devcontainer/devcontainer.json
```diff
@@ -36,7 +36,8 @@
                 "ms-python.python",
                 "ms-python.vscode-pylance",
                 "mohsen1.prettify-json",
-                "redhat.vscode-yaml"
+                "redhat.vscode-yaml",
+                "mintchipleaf.fairyfloss"
```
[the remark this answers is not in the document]
kurysauce: I have been using Ersilia in the Github codespace and I was updating the VSCode interface, if that answers your questions?

ersilia/io/input.py
```diff
@@ -273,4 +273,6 @@ def example(self, n_samples, file_name, simple, try_predefined):
             predefined_done = self.predefined_example(file_name)
         if not predefined_done:
             self.logger.debug("Randomly sampling input")
-            self.random_example(n_samples=n_samples, file_name=file_name, simple=simple)
+            return self.random_example(n_samples=n_samples, file_name=file_name, simple=simple)
+        with open(file_name, "r") as f:
```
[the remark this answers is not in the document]
kurysauce: Thanks for the feedback!

ersilia/publish/test.py
kurysauce: The Mean Relative Absolute Error calculated the mean of the **relative** **absolute** **error**, so I am calculating the relative error between each pair of values and then I take the average of these relative errors across all data points. There did exist standard implementations, but I was running into divide by 0 errors.
Screenshot 2024-08-12 at 2 07 35 PM

However, after testing this implementation on different models, I found that there are high inflated error values that cause the test to fail when it should pass (similar to #1221). So, I am switching to the root mean square error to solve this issue.

kurysauce:
Updated PR @miquelduranfrigola

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
DhanshreeA: Oh okay, I'm not sure why the new package was added in this configuration, since this is the config that gets used to build a codespace environment.

DhanshreeA:
Hi @kurysauce - I've gone through your refactor of the test command and I think this is very much headed in the right direction. I have a small concern - I am unsure if we are accounting for output types within the check for consistent outputs - the code seems to be written only for dealing with numeric output types, as I don't see any check on the output type anywhere. Is that indeed the case?

DhanshreeA:
@kurysauce I've another request for you for this PR. The previous contributors left a lot of print statements within the code, which is generally not a good practice. Would it be possible for you to replace them with debug and info logs? :)

kurysauce:
> @kurysauce I've another request for you for this PR. The previous contributors left a lot of print statements within the code, which is generally not a good practice. Would it be possible for you to replace them with debug and info logs? :)

Sounds good! Just to clarify, what would be appropriate to print vs log? (outputs from failed runs, output from successful runs, indication of new test check, etc...)?

kurysauce:
> Hi @kurysauce - I've gone through your refactor of the test command and I think this is very much headed in the right direction. I have a small concern - I am unsure if we are accounting for output types within the check for consistent outputs - the code seems to be written only for dealing with numeric output types, as I don't see any check on the output type anywhere. Is that indeed the case?

@DhanshreeA I believe the test commands tests `string` as well in these spots: [1](uhttps://github.com/kurysauce/ersilia_test/blob/1beb7558b79dd6a4c9212e963751e472e43d408e/ersilia/publish/test.py#L829C19-L837C27rl) and [2](https://github.com/kurysauce/ersilia_test/blob/1beb7558b79dd6a4c9212e963751e472e43d408e/ersilia/publish/test.py#L465C1-L470C74)

DhanshreeA:
> > @kurysauce I've another request for you for this PR. The previous contributors left a lot of print statements within the code, which is generally not a good practice. Would it be possible for you to replace them with debug and info logs? :)
> 
> Sounds good @DhanshreeA ! Just to clarify, what would be appropriate to print vs log? (outputs from failed runs, output from successful runs, indication of new test check, etc...)?

Generally speaking, there should be no print statements and everything that needs to be logged onto the terminal should essentially be done through a logger and not through print. :)

miquelduranfrigola:
Yes indeed, generally speaking we want to log information instead of printing. Note that logging is activated only with the `-v` flag, though. So, to show on screen print-like messages, we can use the `echo` command from the `click` library as done by the ersilia CLI. 

In any case, there are plenty of examples within the Ersilia codebase as to how we should be logging messages. Usually, they are accessible via the `self.log.debug/info/warning` methods if `ErsiliaBase` is used as a parent class.

I hope this helps!

Issue #977, opened 2024-03-05
[🐅 Epic]: Ersilia Model Hub Maintenance: background and first steps


miquelduranfrigola:
# Background

This is the opening issue of the project [Ersilia Automation Workflow Module](https://docs.google.com/document/d/1dT5ymG8OkxMJ4EbquxLpDpe5bAQ8sAL9/edit?usp=sharing&ouid=114775674178390159004&rtpof=true&sd=true) done in collaboration with [Harvard Tech For Social Good](https://socialgood.hcs.harvard.edu/) (T4SG). The goal is to create suite of GitHub Actions workflows and a Python-based CLI testing module that run several critical monitoring and maintenance tasks for the [Ersilia Model Hub](https://github.com/ersilia-os/ersilia).

* @DhanshreeA and @miquelduranfrigola will be supervising the project from Ersilia's end.
* The main working repository will be [Ersilia Maintenance](https://github.com/ersilia-os/ersilia-maintenance). The [Ersilia CLI](https://github.com/ersilia-os/ersilia) will also be modified.

# Useful resources

To become familiar with the Ersilia Model Hub, these resources can be helpful:
* [Ersilia Book](https://ersilia.gitbook.io): The main documentation of Ersilia can be found in this page.
* [Model contribution pipeline video](https://youtu.be/I7dYI4ZF7Q0?si=CdogDgZum28B6ugj): This recording made by @Lehcar shows how to contribute a model to Ersilia. While model contribution is not the scope of the project, this helps understand the origin of the models.
* [Model catalog in AirTable](https://airtable.com/appgxpCzCDNyGjWc8/shrNc3sTtTA3QeEZu): This spreadsheet shows the current catalog of models, along with their status and metadata.
* [Ersilia Issues](https://github.com/ersilia-os/ersilia/issues): Participating in issue discussions is a great way of becoming familiar with the tool.

# First steps

To get started with the project, we need to:

## Set up users.
- [x] Include T4SG volunteers as collaborators/members to the Ersilia GitHub org. @miquelduranfrigola will create a `t4sg` team in Ersilia.
- [x] Give permissions to submit to the `main` branch of the [ersilia-maintenance](https://github.com/ersilia-os/ersilia-maintenance) repository. @miquelduranfrigola will do it.

## Work on a case example on the Ersilia CLI

[...]

PR #1227, opened 2024-08-13
Move the model inspection code

dzumii:

**Description**

Moving the model inspection code from this [fork](https://github.com/reneeperpignan/ersilia-r) to Ersilia

**Changes to be made**
Implement the inspect command
Implements the ModelInspector class
Register the inspect command within the ersilia cli
Update the inspect_model workflow to actually clone and use ersilia instead
Remove all print calls and replace them with debug logs

**Status**

Implemented the inspect command
Implemented the ModelInspector class
Registered the inspect command within the ersilia cli
Updated the inspect_model workflow to actually clone and use ersilia instead

**To do**

Remove all print calls and replace them with debug logs

Related to #977

### DIFF
--- a/ersilia/cli/commands/inspect.py
+++ b/ersilia/cli/commands/inspect.py
@@ -1,0 +1,30 @@
+from ...publish.inspect import ModelInspector
+import click
+from . import ersilia_cli
+import json
+
+
+def inspect_cmd():
+
+    @ersilia_cli.command(short_help="Inspect model", help="Inspect model structure and metadata")
+    @click.argument("model", type=click.STRING)
+    def inspect(model):
+
+        inspector = ModelInspector(model)
+        value = {
+             'is_github_url_available': inspector.checkRepoExists(0),
+             'is_github_url_available_details': inspector.checkRepoExists(1),
+             'metadata_complete': inspector.metadataComplete(0),
+             'metadata_complete_details': inspector.metadataComplete(1),
+             'folder_structure_complete': inspector.folderStructureComplete(0),
+             'folder_structure_complete_details': inspector.folderStructureComplete(1),
+             'docker_check': inspector.validateDependicies(0),
+             'docker_check_details': inspector.validateDependicies(1),
+             'computational_performance_tracking': inspector.computationalPerformance(0),
+             'computational_performance_tracking_details': inspector.computationalPerformance(1),
+             'extra_files_check': inspector.noExcessFiles(0),
+             'extra_files_check_details': inspector.noExcessFiles(1),
+        }
+        #print or return?
+        print(json.dumps(value))
+        return json.dumps(value)
--- a/ersilia/publish/inspect.py
+++ b/ersilia/publish/inspect.py
@@ -1,0 +1,278 @@
+from .. import ErsiliaBase
+import requests
+import subprocess
+import time
+import json
+from urllib.request import urlopen
+from ..hub.content.card import RepoMetadataFile
+
+
+class ModelInspector(ErsiliaBase):
+    def __init__(self, model, config_json=None):
+        ErsiliaBase.__init__(self, config_json=config_json, credentials_json=None)
+        self.model = model
+
+
+    def checkRepoExists(self, flag):
+        """
+        Verify that the repository exists at a given link.
+        
+        Args:
+            flag (int): Flag indicating whether to return boolean or detailed message.
+            
+        Returns:
+            bool or str: Depending on the flag, returns boolean or a message indicating the check status.
+        """
+        url = f"https://github.com/ersilia-os/{self.model}"
+        response = requests.head(url)
+        if flag == 0:
+            return response.status_code == 200
+        elif response.status_code == 200:
+            return "Check passed."
+        return f"Connection invalid, no github repository found at https://github.com/ersilia-os/{self.model}. Please check that this repository exists in the ersilia-os database."
+    
+    
+    def metadataComplete(self, flag):
+        """
+        Search for specific keys in metadata JSON file.
+        
+        Args:
+            flag (int): Flag indicating whether to return boolean or detailed message.
+            
+        Returns:
+            bool or str: Depending on the flag, returns boolean or a message indicating the check status.
+        """
+        url = f"https://raw.githubusercontent.com/ersilia-os/{self.model}/main/metadata.json" # Get raw file from GitHub
+        if requests.head(url).status_code != 200: # Make sure repo exists
+           if flag == 0:
+                return False
+           return f"Metadata file could not be loated for model {self.model}. Please check that the link https://raw.githubusercontent.com/ersilia-os/{self.model}/main/metadata.json is valid."
+        
+        response = requests.get(url)
+        file = response.json() # Save as json object
+
+        check_passed = True
+        details = ""
+
+        test = RepoMetadataFile(self.model)
+        
+        try:
+            RepoMetadataFile.read_information(test)
+        
+        except Exception as e:
+            details = details + f"Error encountered when parsing through metadata file: {e} "
+            check_passed = False
+
+        if file is not None:
+            try:
+                if file['Publication'] and file['Source Code'] and file['S3'] and file['DockerHub']: # Parse through json object and ensure 
+                    
+                    pub_url_works = requests.head(file['Publication']).status_code != 404
+                    if not pub_url_works:
+                        details = details + f"The URL ({file['Publication']}) listed in the publication field of the metadata was not found (Error code {requests.head(file['Publication']).status_code}). Please verify that this link is accurate. "
+                        check_passed = False
+                    
+                    source_url_works = requests.head(file['Source Code']).status_code != 404
+                    if not source_url_works:
+                        details = details + f"The URL ({file['Source Code']}) listed in the source code field of the metadata was not able to be accessed (Error code {requests.head(file['Source Code']).status_code}). Please verify that this link is accurate. "
+                        check_passed = False
+                    
+                    s3_url_works = requests.head(file['S3']).status_code != 404
+                    if not s3_url_works:
+                        details = details + f"The URL ({file['S3']}) listed in the S3 field of the metadata was not able to be accessed (Error code {requests.head(file['S3']).status_code}). Please verify that this link is accurate. "
+                        check_passed = False
+
+                    docker_url_works = requests.head(file['DockerHub']).status_code != 404
+                    if not docker_url_works:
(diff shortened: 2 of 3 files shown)


DhanshreeA:
@dzumii this looks good to me, but I realised that we need to change the function names from camelCase to snake_case, especially within the ModelInspector class. Could you please make that change and then we can merge this PR?

dzumii:
Ok, I will do that, and also change print calls to debug logs, so you can merge

ersilia/publish/inspect.py
```diff
@@ -0,0 +1,378 @@
...
+            )
+            check_passed = False
+
+        if flag == 0:
+            return check_passed
+        if check_passed:
+            return "Check passed."
+        return details
+
+    def computational_performance(self, flag):
```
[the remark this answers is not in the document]
dzumii: Done

ersilia/cli/commands/inspect.py
```diff
@@ -0,0 +1,44 @@
...
+            "is_github_url_available": inspector.check_repo_exists(0),
+            "is_github_url_available_details": inspector.check_repo_exists(1),
+            "metadata_complete": inspector.metadata_complete(0),
+            "metadata_complete_details": inspector.metadata_complete(1),
+            "folder_structure_complete": inspector.folder_structure_complete(0),
+            "folder_structure_complete_details": inspector.folder_structure_complete(1),
+            "docker_check": inspector.validate_dependencies(0),
+            "docker_check_details": inspector.validate_dependencies(1),
+            "computational_performance_tracking": inspector.computational_performance(
+                0
```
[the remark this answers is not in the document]
dzumii: I am waiting for your decision here @DhanshreeA @miquelduranfrigola. I think we can remove the `computational_performance_tracking` function and leave the`computational_performance_tracking_details` only

dzumii:
@DhanshreeA the computational_performance_tracking function now works well, see the result of the inspect command
[here](https://github.com/user-attachments/files/16635492/inspect_output.log)

ersilia/publish/inspect.py
```diff
@@ -0,0 +1,359 @@
...
+            bool or str: Depending on the verbose, returns boolean or a message indicating the check status.
+        """
+        details = ""
+
+        for n in (1, 10, 100):
+
+            cmd = (
+                f"ersilia serve {self.model} && "
+                f"ersilia example -f my_input.csv -n {n} && "
+                "ersilia run -i my_input.csv"
```
DhanshreeA, line 287: Hey @dzumii please also add a command to close ersilia here otherwise we will end up serving the same model multiple times.

DhanshreeA (changes requested):

ersilia/publish/inspect.py
```diff
@@ -0,0 +1,359 @@
...
+        """
+        url = f"https://github.com/ersilia-os/{self.model}"
+        response = requests.head(url)
+        if flag == 0:
+            return response.status_code == 200
+        elif response.status_code == 200:
+            return "Check passed."
+        return f"Connection invalid, no github repository found at https://github.com/ersilia-os/{self.model}. Please check that this repository exists in the ersilia-os database."
+
+    def metadata_complete(self, flag):
```
DhanshreeA, line 33: We can rename this to `check_complete_metadata`

ersilia/publish/inspect.py
```diff
@@ -0,0 +1,359 @@
...
+                )
+
+        if flag == 0:
+            return check_passed
+        if details:
+            return details
+        return "Check passed."
+
+    def folder_structure_complete(self, flag):
+        """
```
DhanshreeA, line 141: Rename this method to `check_complete_folder_structure`

ersilia/publish/inspect.py
```diff
@@ -0,0 +1,359 @@
...
+                )
+                check_passed = False  # If the folder URL is not valid return false
+
+        if flag == 0:
+            return check_passed
+        if check_passed:
+            return "Check passed."
+        return details
+
+    def validate_dependencies(self, flag):
```
DhanshreeA, line 194: Rename this to `check_dependencies_are_valid`

ersilia/publish/inspect.py
```diff
@@ -0,0 +1,359 @@
...
+            check_passed = False
+
+        if flag == 0:
+            return check_passed
+        if check_passed:
+            return "Check passed."
+        return details
+
+    def computational_performance(self, verbose):
+        """
```
DhanshreeA, line 271: Rename this method to `check_comptuational_performance`

ersilia/publish/inspect.py
```diff
@@ -0,0 +1,359 @@
...
+            endTime = time.time()
+
+            executionTime = endTime - startTime
+            details += f"Execution time ({n} Prediction(s)): {executionTime} seconds. "
+
+        if verbose == 1:
+            return details
+        return True
+
+    def no_excess_files(self, flag):
```
DhanshreeA, line 314: Rename this to `check_no_extra_files`

DhanshreeA (changes requested):

DhanshreeA:
@dzumii let's take a call on this - So generally, this style of implementation is very bad since it runs the same function twice only to use partial results from it in each call, thereby making the whole implementation slow. 

```python
    def inspect(self, model):
        inspector = ModelInspector(model)
        value = {
            "is_github_url_available": inspector.check_repo_exists(0),
            "is_github_url_available_details": inspector.check_repo_exists(1),
            "metadata_complete": inspector.metadata_complete(0),
            "metadata_complete_details": inspector.metadata_complete(1),
            "folder_structure_complete": inspector.folder_structure_complete(0),
            "folder_structure_complete_details": inspector.folder_structure_complete(1),
            "docker_check": inspector.validate_dependencies(0),
....
```

And as we can see it's happening for all the checks implemented in `Model Inspector`. We need an overhaul on this one, and this will happen in two places, in the `inspect` function in `ersilia/cli/commands/inspect.py`, and within each of the checks in the `ModelInspector` class. 

Let's start with the `ModelInspector` class:
1. We should remove the `flag` variable from each of the `check_xxxx` methods since it does nothing useful other than returning whether the check was successful or not. 
2. We should return the output from each of the `check_xxxx` methods in a tuple of the form `(boolean, str)` where the first element in the tuple is the success status of the check, ie True for it succeeded and False for failure, and then the actual str which gives details from the check.


In the `InspectCommand` implementation, we should modify the checks to something like:

```python

    def inspect(self, model):
        inspector = ModelInspector(model)
        check_repo_exists_result = inspector.check_repo_exists()
        value = {
            "is_github_url_available": check_repo_exists_result[0],
            "is_github_url_available_details": check_repo_exists_response[1],
.....
```

ersilia/cli/commands/inspect.py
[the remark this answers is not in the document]
DhanshreeA: Hey @dzumii refer to my comment [here](https://github.com/ersilia-os/ersilia/pull/1227#issuecomment-2295844688). I think we need to overhaul this entire implementation.

DhanshreeA:
> @DhanshreeA the computational_performance_tracking function now works well, see the result of the inspect command [here](https://github.com/user-attachments/files/16635492/inspect_output.log)

The result looks good @dzumii !

dzumii:
> @dzumii let's take a call on this - So generally, this style of implementation is very bad since it runs the same function twice only to use partial results from it in each call, thereby making the whole implementation slow.
> 
> ```python
>     def inspect(self, model):
>         inspector = ModelInspector(model)
>         value = {
>             "is_github_url_available": inspector.check_repo_exists(0),
>             "is_github_url_available_details": inspector.check_repo_exists(1),
>             "metadata_complete": inspector.metadata_complete(0),
>             "metadata_complete_details": inspector.metadata_complete(1),
>             "folder_structure_complete": inspector.folder_structure_complete(0),
>             "folder_structure_complete_details": inspector.folder_structure_complete(1),
>             "docker_check": inspector.validate_dependencies(0),
> ....
> ```
> 
> And as we can see it's happening for all the checks implemented in `Model Inspector`. We need an overhaul on this one, and this will happen in two places, in the `inspect` function in `ersilia/cli/commands/inspect.py`, and within each of the checks in the `ModelInspector` class.
> 
> Let's start with the `ModelInspector` class:
> 
> 1. We should remove the `flag` variable from each of the `check_xxxx` methods since it does nothing useful other than returning whether the check was successful or not.
> 2. We should return the output from each of the `check_xxxx` methods in a tuple of the form `(boolean, str)` where the first element in the tuple is the success status of the check, ie True for it succeeded and False for failure, and then the actual str which gives details from the check. In this you can even use a [NamedTuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) instead of a simple tuple (I personally like named tuples instead of the primitive tuples)
> 
> In the `InspectCommand` implementation, we should modify the checks to something like this if you use primitive tuple:
> 
> ```python
>     def inspect(self, model):
>         inspector = ModelInspector(model)
>         check_repo_exists_result = inspector.check_repo_exists()
>         value = {
>             "is_github_url_available": check_repo_exists_result[0],
>             "is_github_url_available_details": check_repo_exists_response[1],
> .....
> ```
> 
> Or something like this if you use named tuple:
> 
> ```python
>     def inspect(self, model):
>         inspector = ModelInspector(model)
>         check_repo_exists_result = inspector.check_repo_exists()
>         value = {
>             "is_github_url_available": check_repo_exists_result.status,
>             "is_github_url_available_details": check_repo_exists_response.detail,
> .....
> ```

Alright @DhanshreeA, a call will do

miquelduranfrigola:
Hi @DhanshreeA - feel free to merge this as you see fit.

ersilia/publish/inspect.py
dzumii: done

ersilia/publish/inspect.py
dzumii: done

ersilia/publish/inspect.py
dzumii: done

ersilia/publish/inspect.py
dzumii: done

ersilia/publish/inspect.py
dzumii: done

ersilia/publish/inspect.py
dzumii: done

dzumii:
@DhanshreeA I have implemented the changes you suggested, i tried it out, see the output 
[here](https://github.com/user-attachments/files/16675097/inspect_output2.log)

DhanshreeA (approved):

DhanshreeA:
LGTM @dzumii Thank you!

dzumii:
You are welcome @DhanshreeA

PR #1300, opened 2024-10-07
Update approve-trigger.yml

### LINKED ISSUE #962: 🐛 Bug: `/approve` command triggers workflows outside the "Model Request" issue template
### Describe the bug.

The `/approve` command triggers workflows to create a model repository when an Ersilia Maintainer approves a model request issue. This workflow runs every time a user (possibly only Ersilia maintainers) comment on any issue within this repository, whereas this workflow should be triggered only when the related issue is a "Model Request" issue.

This is a good first issue and provides a starting point for learning more about GitHub actions. The workflow being referenced is implemented [here](https://github.com/ersilia-os/ersilia/blob/master/.github/workflows/approve-trigger.yml). 

This can potentially be fixed by making the job within this workflow execute conditionally based on the issue label. You can find an example [here](https://github.com/ersilia-os/ersilia-self-service/blob/2d8f19d4532357743c7fb6ccf319e7d92a78e84c/.github/workflows/self-service.yml#L8).

You can see the number of times this action has run where it was not relevant [here](https://github.com/ersilia-os/ersilia/actions/workflows/approve-trigger.yml). 

To reproduce this issue yourself, please open an issue, and tag one of the maintainers to comment on it. Once a maintainer comments on it, the workflow will be triggered.

### Expected behavior.

One possibility would be to modify the first triggered action to check for a `new-model`, as suggested by @GrantBirki . 

To make sure the implemented fix works, please tag a maintainer to post a comment on an issue and verify if the workflow no longer gets triggered. 

### Screenshots.

_No response_

### Operating environment

Ubuntu 20.04 LTS

### Additional context

_No response_

aderemi1224:

**Description**
This project addresses the current implementation of the /approve command in the Ersilia repository, which unintentionally triggers workflows to create a model repository for any comment made on any issue. The objective is to refine this functionality so that the workflow is executed only for "new-model" issues. Originally, the workflow would activate for any comment on any issue, without checking if the issue was related to creating a new model. This led to unwanted triggers.

**Changes to be made**
* I updated the if condition in the job approve-command-trigger to ensure the workflow only runs when the issue has a label matching 'new model':
        if: contains(github.event.issue.labels.*.name, 'new-model')
This ensures that the workflow does not activate for every comment but is instead filtered to respond only when a specific label is applied to the issue.
    

**Status**
- I have successfully modified the trigger to activate only when comments are made on issues labeled "new-model". The label-based filtering using the if condition ensures that the workflow now only executes when an issue has the "new-model" label, preventing it from triggering other types of issues.

Closes #962


aderemi1224:
@DhanshreeA Please kindly review

dzumii:
@aderemi1224 Have you tested your changes? If yes, can you share details with us regarding your test before and after the changes were made?

.github/workflows/approve-trigger.yml
```diff
@@ -2,16 +2,25 @@ name: Approve Command Trigger
...
-      - name: Approve Command Trigger
-        uses: peter-evans/slash-command-dispatch@13bc09769d122a64f75aa5037256f6f2d78be8c4 # pin@v4.0.0
-        with:
-          token: ${{ secrets.PAT }}
-          issue-type: issue
-          permission: write
-          reactions: false
-          commands: |
-            approve
+    - name: Log Checking for 'Model Request' label
```
DhanshreeA, line 10: We don't really need these logs. We can remove this.

.github/workflows/approve-trigger.yml
```diff
@@ -2,16 +2,25 @@ name: Approve Command Trigger
...
+      uses: peter-evans/slash-command-dispatch@13bc09769d122a64f75aa5037256f6f2d78be8c4 # pin@v4.0.0
+      with:
+        token: ${{ secrets.PAT }}
+        issue-type: issue
+        permission: write
+        reactions: false
+        commands: |
+          approve
+
+    - name: Log Approval process completed
```
DhanshreeA, line 24: Same as above, we don't need this.

.github/workflows/approve-trigger.yml
```diff
@@ -2,16 +2,25 @@ name: Approve Command Trigger
 on:
   issue_comment:
     types: [created]
+
 jobs:
   approve-command-trigger:
```
DhanshreeA, line 7: We can add `if` conditions at the level of the job itself. For example, you can implement the conditional check on the issue's label as done here: https://github.com/ersilia-os/ersilia-self-service/blob/bd9203cc3370bd405d11dc99b09737c147c5223e/.github/workflows/self-service.yml#L8

DhanshreeA (changes requested):

.github/workflows/approve-trigger.yml
aderemi1224: @DhanshreeA Okay. Will make these changes. Thank you

.github/workflows/approve-trigger.yml
```diff
@@ -2,16 +2,18 @@ name: Approve Command Trigger
...
-      - name: Approve Command Trigger
-        uses: peter-evans/slash-command-dispatch@13bc09769d122a64f75aa5037256f6f2d78be8c4 # pin@v4.0.0
-        with:
-          token: ${{ secrets.PAT }}
-          issue-type: issue
-          permission: write
-          reactions: false
-          commands: |
-            approve
+    - name: Approve Command Trigger
```
DhanshreeA, line 11: We should keep the single tab indentation

DhanshreeA (changes requested):

.github/workflows/approve-trigger.yml
```diff
@@ -2,16 +2,18 @@ name: Approve Command Trigger
 on:
   issue_comment:
     types: [created]
+
 jobs:
   approve-command-trigger:
+    if: contains(github.event.issue.labels.*.name, 'Model Request')
```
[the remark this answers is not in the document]
DhanshreeA: Refer here, for example: https://github.com/ersilia-os/ersilia/issues?q=is%3Aopen+is%3Aissue+label%3Anew-model

.github/workflows/approve-trigger.yml
DhanshreeA: @aderemi1224 can you please restore the indentation?

Aderemi24:
@DhanshreeA I have made changes

Issue #925, opened 2023-12-21
GitHub Action to automatically test the `--track` flag in a model run


miquelduranfrigola:
The `ersilia run --track` flag uploads a model digest to an S3 bucket that is eventually monitored by Splunk. It would be good to have a `cron` job that does a short run to ensure that things are up-and-running. Suggestions are welcome!

GemmaTuron:
Hi @DhanshreeA 

You marked this as Ready, but the issue is not closed. what is the status?

DhanshreeA:
This is not ready and I marked that incorrectly. Currently the --track flag is not incorporated anywhere other than [ersilia-self-service](https://github.com/ersilia-os/ersilia-self-service/blob/2d8f19d4532357743c7fb6ccf319e7d92a78e84c/.github/workflows/self-service.yml#L46).

However, I think we should bundle this with the `ersilia inspect` command since that command actually runs the same model a couple of times, for example, with 1, 10, 100 inputs - and an approach like that would be better suited for tracking the computational requirements of an individual model; and collecting the similar data to run a comparison across models. 

At present within tracking, we're just taking an average of the memory and CPU percentage consumed regardless of the input sizes - but with that information incorporated, our analyses would be more substantial. So to conclude, we should bundle track with ersilia inspect, and potentially run this within `ersilia maintenance` repository.

GemmaTuron:
Hi @DhanshreeA 

This is a Batch initiative. Can you detail here the tasks to be completed for this, so I can convert them to issues (maybe in the ersilia-maintenance repository).

DhanshreeA:
I'll make sure to answer this before EOD today.

DhanshreeA:
Go ahead please! @AnushkaShreyam

AnushkaShreyam:
Hi @DhanshreeA,
I hope you're doing well. Could you please kindly review the other pull request when you have a moment? Also, I noticed my name isn't listed as the assignee for this particular issue. If there's anything I need to do to address that, please let me know. Thank you so much for your guidance!

PR #1327, opened 2024-10-17
Update inspect.py to include --track flag support for model inspection

AnushkaShreyam:
This pull request updates the inspect.py script to improve model inspection by adding support for the --track flag. This change allows users to enable detailed tracking information during model inspection.

Key Changes:

    Added --track flag: The script now accepts the --track flag as an optional argument, which will provide extended logging and performance tracking during the inspection process.
    Enhanced logging: The logging functionality has been updated to capture and display tracking details when the --track flag is active.
    Code refactoring: Minor refactoring to ensure consistent handling of input arguments and improved readability of the script.

Related to #925

### DIFF
--- a/ersilia/publish/inspect.py
+++ b/ersilia/publish/inspect.py
@@ -257,21 +257,22 @@ class ModelInspector(ErsiliaBase):
             return Result(True, "Check passed.")
         return Result(False, details)
 
+
     def check_comptuational_performance(self):
         """
         Measure computational performance by serving the model and running predictions.
+        Includes tracking metrics using the --track flag.
 
-         Returns:
+        Returns:
             Result: A namedtuple containing a boolean success status and details of the check.
         """
         details = ""
 
         for n in (1, 10, 100):
-
             cmd = (
                 f"ersilia serve {self.model} && "
                 f"ersilia example -f my_input.csv -n {n} && "
-                "ersilia run -i my_input.csv && "
+                "ersilia run -i my_input.csv --track && "  # Added --track flag
                 "ersilia close"
             )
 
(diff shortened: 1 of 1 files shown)


ersilia/publish/inspect.py
```diff
@@ -257,21 +257,22 @@ def check_dependencies_are_valid(self):
...
         """
         details = ""
 
         for n in (1, 10, 100):
-
             cmd = (
                 f"ersilia serve {self.model} && "
                 f"ersilia example -f my_input.csv -n {n} && "
-                "ersilia run -i my_input.csv && "
+                "ersilia run -i my_input.csv --track && "  # Added --track flag
```
DhanshreeA, line 275: @AnushkaShreyam track is used when serving a model, so typically, you would want to do something like `ersilia serve eosxxxx --track`.

ersilia/publish/inspect.py
```diff
@@ -295,6 +296,16 @@ def check_comptuational_performance(self):
...
+            # Try to read and analyze tracking data
+            try:
+                tracking_file = sorted(self.root.joinpath('.ersilia/tracking').glob('*.json'))[-1]
+                with open(tracking_file) as f:
+                    tracking_data = json.load(f)
+                details += f"Memory usage ({n} predictions): {tracking_data.get('memory_percent', 'N/A')}%. "
+                details += f"CPU usage ({n} predictions): {tracking_data.get('cpu_percent', 'N/A')}%. "
+            except (IndexError, FileNotFoundError, json.JSONDecodeError) as e:
+                details += f"Warning: Could not read tracking data for {n} predictions. "
+
```
DhanshreeA, line 308: I don't understand what's happening here. We don't really need this section at all because of the way that tracking is implemented. Once a model is served with the track flag, tracking data is collated at the end of a run and simply uploaded to an S3 bucket. It is not saved in a file like tracking.json that you are using here.

DhanshreeA (approved):

Issue #1299, opened 2024-10-07
🐕 Batch: Ersilia CLI Performance Improvement


DhanshreeA:
### Summary

@Abellegese is looking into improving the design architecture of core components within the Ersilia CLI, as well as replacing sequential execution with asynchronous approaches. 
On top of this this batch will be used to report results from benchmarking and profiling the current implementation, with reporting updates as we improve the architecture and implement asynchronous execution within flows.

Abellegese:
The first objectives to improve performance on the Ersilia CLI are summarized below:

1) Improving performance of fetch command.
    -Looking for unnecessary  check ups, high latency interfaces, and asynchronizeable IOs
2) Improving performance of serve command.
    -On the preliminary level, this does not seem to have issues but I will look into ways to make it light (function stack calling is crazy everywhere).
3) Improving performance of run command.
    -This probably the most important CLI that needs performance improvement (speed mainly).
        -This command would be benefited from concurrency, async ops, high speed module binding and simplified 
         preprocessing steps.

The second objective is to improve performance on the IO operations on the global level.

1) Defining extensive IOs as async and concurency ops
2) Registering common IOs as singleton
3) Allowing template based Aync Ops creation with method specified on 2)
4) Module binding on ersilia command call is appeared to be some performance overhead, so making this fast on the global level


....Will continue as move on.

Abellegese:
## Preliminary Report

**Fetch CLI**

I analyzed this command using **[py-spy and speedscope.app]**. The results shows that this command has a ton of unnecessary (**costly**) checkups.

**1)**  As @DhanshreeA  mentioned on issue #1259, the decision flow for model fetching appears to have high performance bottleneck.
      - I checked the network latency between the dockerhub and airtable during checkup and fetching, the airtable had 
         high network latency. 

This is the profiling results for original decision flow when model fetching (CPU time).
![original_deicision_flows](https://github.com/user-attachments/assets/4d57a249-2be3-4b01-9592-c95d10d28579)

And this is the simplified decision flow I tried to see how it goes.
![simplified_decision_flows](https://github.com/user-attachments/assets/f4480523-76f4-4083-8ca2-929b2bd9defa)

**Opinions**
------------------
In my opinion, I would use prior knowledge(Where each models are located and which models repos has high bandwidth or low latency, which was docker in my case) about the model repository and use that information and 
[...]

PR #1343, opened 2024-10-23
Performance Improvement on the Airtable Interface(#1299)

Abellegese:

**Description**

A few improvement made on the airtable interface as listed below:

1) Changing manual yield with more efficient and auto sub-generator creator yield from
In terms of performance, yield from can be more efficient than manually using yield with a loop. This is because yield from is implemented in C and optimized to handle delegation more efficiently and also has memory improvement. Will report time comparison on the #1299.

2) Using pyairtable is inevitable so instead of installing it during AirtableInterface call, we can install it as requirements(in pyproject.toml), that reduce the AirtableInterface overhead.

Related to #1299


miquelduranfrigola:
Thanks @Abellegese 
I am agree with the performance increase and it is a good suggestion. Thanks.
As for the installation of `pyairtable`, I am less sure. We eventually want to get rid of it. @DhanshreeA can speak more about it. I'll leave it up to both of you to decide how to proceed. At a high level, the way I see it is that AirtableInterface should be a "last resource" option, but definitely not the front-line option.

Abellegese:
Yes @miquelduranfrigola, on the first pull that you merged, I improve 120,000x time the json based model searching functions so this one is just an alternative. Thanks.

DhanshreeA (changes requested):

PR #1344, opened 2024-10-23
Performace Improvement on the run CLI(ersilia-os#1299)

Abellegese:

**Description**

Overall performance of the run CLI has been increased substantially by making the changes below(comparison results will be given at #1299)

1) Asynchronous and concurrent based key converter at the compound identifier
2) Caching(with dynamic custom cache size) key conversion results for smiles both from the APIs and from rdkit **(main performance boost)**
3) Adding a valid smiles checker
4) api schema and information file reader function improvement
5) reduction in the unnecessary heavy app layers
6) more

Related to #1299


Abellegese:
For some reason the pytest ```test_models.py``` test keep failing with an exception ```No Fetcher Available```. Also in this test case the ```predic``` function has been called from the ```ErsiliaModel``` which does not exist. Those two things blocked me for while I did some fixes in the ```test_models.py```:

1) Remove the ```predict``` function
2) Changing the ```repo_path``` -> ```overwrite=True```
3) Changing the model id from ```eos0t02``` to ```eos3b5e```

This is the only solution that worked for me. You can take a look and decide @DhanshreeA and @miquelduranfrigola.

Issue #1370, opened 2024-11-08
🐈 Task: Remove duplicate flag --repo_path from the fetch command


DhanshreeA:
### Summary

Refer to #1262 

The `--repo_path`  flag is a duplicate of the `--from_dir` flag since both expect the local path of a model's directory (repository). Currently the `--from_dir` flag is used only in [setting](https://github.com/ersilia-os/ersilia/blob/bb202b261bb6d4bbe658d207aba7f0c3de495787/ersilia/hub/fetch/fetch.py#L76) the model source, and is not really used elsewhere. Users and contributors mainly use the `repo_path` flag to fetch a locally available model, however for the sake of consistency across nomenclature, we should favor keeping `--from_dir` instead of `--repo_path`, since that is how the other source flags are specified (eg --from_github, --from_dockerhub etc)

### Objective(s)

- [x] Remove the [`--repo_path`](https://github.com/ersilia-os/ersilia/blob/bb202b261bb6d4bbe658d207aba7f0c3de495787/ersilia/cli/commands/fetch.py#L25) flag in favor of only keeping the `--from_dir` flag. 
- [x] For sake of backward compatibility and not breaking the rest of the code, we can set [`repo_path`](https://github.com/ersilia-os/ersilia/blob/bb202b261bb6d4bbe658d207aba7f0c3de495787/ersilia/cli/commands/fetch.py#L108) argument in the `ModelFetcher` class's constructor with the value the user inputs through `from_dir`.

snufkinwa:
Hi, I'd like to contribute to this issue if it's still available.

DhanshreeA:
Hi @snufkinwa thank you for your interest in working on this. We are still finalizing whether we want to address this issue or not.

DhanshreeA:
@miquelduranfrigola @GemmaTuron what do you both think? I know we've been using `repo_path` from quite some time now but it was also developed in the early days of ersilia when we hadn't formalized the CLI to the extent that exists now. So I vote removing this flag and actually using the `from_dir` flag.

snufkinwa:
Let me know! If there are any other first-time issues I can contribute to, I'd love to help.

DhanshreeA:
@snufkinwa I have discussed with @miquelduranfrigola - we do indeed want to get rid of this. If you're still interested in taking this up, I'll be happy to provide support.

miquelduranfrigola:
Thanks @DhanshreeA 

[...]

PR #1381, opened 2024-11-14
Remove duplicate flag --repo_path from the fetch command (#1370)

snufkinwa:

**Description**

The `--repo_path` and `--from_dir` flags both serve the same purpose: specifying the local path to a model's directory. Right now, `--from_dir` is only used to set the model source, while most users rely on `--repo_path` to load local models. To keep naming consistent with other flags (like `--from_github` and `--from_dockerhub`), we should use `--from_dir` and remove `--repo_path`.

**Changes to be made**

- [x] Remove the [`--repo_path`](https://github.com/ersilia-os/ersilia/blob/bb202b261bb6d4bbe658d207aba7f0c3de495787/ersilia/cli/commands/fetch.py#L25) flag in favor of only keeping the --`from_dir` flag.
- [x] For sake of backward compatibility and not breaking the rest of the code, we can set [`repo_path`](https://github.com/ersilia-os/ersilia/blob/bb202b261bb6d4bbe658d207aba7f0c3de495787/ersilia/cli/commands/fetch.py#L108) argument in the `ModelFetcher` class's constructor with the value the user inputs through `from_dir`.

**Status**

Ready to Review

**To do**

- [x] Implementing test for changes, making sure the `fetch` command is working properly 
```python
import unittest
from ersilia.hub.fetch.fetch import ModelFetcher

class TestModelFetcher(unittest.TestCase):
    
    def test_from_dir_priority_when_repo_path_is_set(self):
        fetcher = ModelFetcher(from_dir="test_repo_path")
        self.assertEqual(fetcher.repo_path, "test_repo_path", "repo_path should match the value provided to from_dir")

    def test_from_dir_fallback_when_repo_path_is_none(self):
        fetcher = ModelFetcher(from_dir="fallback_dir")
        self.assertEqual(fetcher.repo_path, "fallback_dir", "from_dir should be used as the primary path")

    def test_no_repo_path_or_from_dir(self):
        fetcher = ModelFetcher(from_dir=None)
        self.assertIsNone(fetcher.repo_path, "repo_path should be None if from_dir is None")

if __name__ == '__main__':
    unittest.main()
```

Related to #1370


snufkinwa:
Removing `@click.option("--repo_path", "-r", default=None, type=click.STRING)` causes the code to break. To fix this, we can use the `or` operator in the `ModelFetcher` constructor to prioritize `from_dir` as the replacement for `repo_path`.

On my Mac, CMake is missing, so while the `fetch` command works to retrieve the model, the build process fails due to the absence of CMake. I ran `test_models.py`, that successfully completed without issues. Additionally, I wrote a few unit tests to ensure `--from_dir` is prioritized correctly in the absence of `repo_path`. 

Screenshot 2024-11-14 at 3 06 35 PM

ersilia/cli/commands/fetch.py
```diff
@@ -105,7 +105,7 @@ def fetch(
             fg="blue",
         )
         mf = ModelFetcher(
-            repo_path=repo_path,
+            from_dir=repo_path or from_dir,
```
DhanshreeA, line 108: I would not do this. I would keep the repo_path parameter in the `ModelFetcher` class, and set it with the value of `from_dir`.

ersilia/hub/fetch/fetch.py
```diff
@@ -22,7 +22,7 @@ def __init__(
         config_json=None,
         credentials_json=None,
         overwrite=None,
-        repo_path=None,
+        from_dir=None,
```
DhanshreeA, line 25: I would recommend using the value of `from_dir` from the CLI to set both `from_dir`, and `repo_path` parameters for the `ModelFetcher` constructor.

DhanshreeA (changes requested):

DhanshreeA:
@snufkinwa we actually need to remove the [flag](https://github.com/ersilia-os/ersilia/blob/92823da32f9592c4c8dc06ee468711293ddd913f/ersilia/cli/commands/fetch.py#L25) from the CLI which I don't see being addressed in this PR.

snufkinwa:
Everything on `/hub/fetch/fetch.py` was reverted back, that was the reason I was getting error when trying to remove `--repo_path`, then add `from_dir` to the constructor.  The only file edited was `cli/commands/fetch.py`. Tested several times with `test_models.py` It fetched successfully.

snufkinwa:
```
  File "/home/runner/work/ersilia/ersilia/ersilia/cli/commands/__init__.py", line 22, in wrapper
    return func(*args, **kwargs)
TypeError: fetch_cmd.<locals>.fetch() missing 1 required positional argument: 'repo_path'
```

This is the error I was getting originally, this happens with the removal of  `@click.option("--repo_path", "-r", default=None, type=click.STRING) `

The `test_models.py` was tested several times, it didn't give me that error when fetching, but the additional test on the PR produced the same error. It has something to do with `BentoML`, as the line in question is in function `def bentoml_common_params(func)`

DhanshreeA:
> ```
>   File "/home/runner/work/ersilia/ersilia/ersilia/cli/commands/__init__.py", line 22, in wrapper
>     return func(*args, **kwargs)
> TypeError: fetch_cmd.<locals>.fetch() missing 1 required positional argument: 'repo_path'
> ```
> 
> This is the error I was getting originally, this happens with the removal of `@click.option("--repo_path", "-r", default=None, type=click.STRING) `
> 
> The `test_models.py` was tested several times, it didn't give me that error when fetching, but the additional test on the PR produced the same error. It has something to do with `BentoML`, as the line in question is in function `def bentoml_common_params(func)`

@snufkinwa this happens because we still need to remove `repo_path` from the arguments that the `fetch` function gets [here](https://github.com/ersilia-os/ersilia/blob/c4c790a20a0e85f07b3f4e7603559e30caba67b2/ersilia/cli/commands/fetch.py#L80).

DhanshreeA (approved):

DhanshreeA:
LGTM @snufkinwa thanks!

PR #1397, opened 2024-11-23
Pytest CLI test using partial mocking

Abellegese:

**Description**

Upgrade the current pytest functionality by including the testing the five CLI using a partial mocking technique.


test/cli/test_catalog.py
```diff
@@ -0,0 +1,59 @@
...
+from unittest.mock import patch, MagicMock
+from click.testing import CliRunner
+from ersilia.cli.commands.catalog import catalog_cmd
+
+@pytest.fixture
+def runner():
+    return CliRunner()
+
+@patch("ersilia.hub.content.catalog.ModelCatalog.airtable", return_value=None)
+@patch("ersilia.hub.content.catalog.ModelCatalog.local")
```
DhanshreeA, line 11: We shouldn't be mocking this because it's not an external interface, but information available on the disk.

test/cli/test_catalog.py
```diff
@@ -0,0 +1,59 @@
...
+@patch("ersilia.hub.content.catalog.ModelCatalog.hub")
+@patch("ersilia.hub.content.card.ModelCard.get", return_value={"model": "metadata"})
```
DhanshreeA, line 13: This information is present on the disk as well, especially for the models available locally.

test/cli/test_catalog.py
```diff
@@ -0,0 +1,59 @@
...
+    elif "--card" in options and model: 
+        mock_model_card_get.assert_called_once()
+    elif "--hub" in options:
+        mock_model_catalog_hub.assert_called_once()
+    elif "--browser" in options:
+        mock_model_catalog_airtable.assert_called_once()
+    elif "--hub" in options:
+        mock_model_catalog_hub.assert_called_once()
+    elif "--local" in options:
+        mock_model_catalog_local.assert_called_once()
```
DhanshreeA, line 55: We can also add a test for an exception being raised if both `--hub`, and `--local` are present in options bec both shouldn't be specified together.

test/cli/test_delete.py
```diff
@@ -0,0 +1,47 @@
...
+from ersilia.cli.commands.delete import delete_cmd
+from ersilia.utils.logging import logger
+
+@pytest.fixture
+def runner():
+    return CliRunner()
+
+MODEL_ID = "eos3b5e"
+@patch("ersilia.core.modelbase.ModelBase")
+@patch("ersilia.hub.delete.delete.ModelFullDeleter.needs_delete", return_value=None)
```
DhanshreeA, line 13: We should divide this into two functions: when the model is available, where the return value from this mock would be 'True', and then the other case where model is not available, making the return value from the mock 'False', which would change the output message from the mock for the assertion.

DhanshreeA (changes requested):

test/cli/test_catalog.py
Abellegese: Solved, now local can read models that are written in `eos` folder. Nice catch @DhanshreeA.

test/cli/test_delete.py
Abellegese: Resolved, but found some bug @DhanshreeA , that even the model is not there it consider it as it existed and requires deletion (this function `needs_delete`). So for now could not create the second function. But this current change the general health of this `cmd`.

PR #1403, opened 2024-11-26
Bentoml loop fix

Abellegese:

**Description**

Installing bentoml at installation time and removed the checkup.


DhanshreeA:
Hi @Abellegese thank you for the PR, however this is not how we want to address this since we don't want to make a hard requirement out of bentoml by adding it to the pyproject.toml. I've pushed a [fix](https://github.com/ersilia-os/ersilia/commit/a780714d0100cab834b639de652d74d6c9f4e6f8) here which should make the loop not happen because the version file isn't temporary anymore.

PR #1411, opened 2024-12-04
Generalize Standard Run 

DhanshreeA:
The aim of this PR is to generalize standard run to work with any type of model server response and be able to serialize a CSV output file. In addition to that, other changes introduced through this PR are:

- We can now calculate the output csv header from a predefined example file, in addition to the standard csv output file
- The is_ready function from the StandardCSVRunApi class is removed, and only the is_amenable function is kept because the checks in the latter were a superset of the ones in is_ready.
- Unused code is removed
- The `run` method no longer has a try_standard parameter because standard run is the default behavior.

### DIFF
--- a/ersilia/cli/commands/run.py
+++ b/ersilia/cli/commands/run.py
@@ -24,17 +24,11 @@ def run_cmd():
         "-b", "--batch_size", "batch_size", required=False, default=100, type=click.INT
     )
     @click.option(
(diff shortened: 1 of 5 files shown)


Abellegese (approved):
Seems solid @DhanshreeA, everything makes sense. I think we can merge it.

PR #1485, opened 2025-01-03
change minimum length for description to 10

miquelduranfrigola:
I have simply changed the minimum description length for the columns
Related to issue for model [eos4djh](https://github.com/ersilia-os/eos4djh/issues/3)


DhanshreeA:
@miquelduranfrigola I don't understand why we want to do this though? Isn't 10 words too short for a description?

miquelduranfrigola:
Not really, @DhanshreeA. This is for the columns description, not the model description. Columns like "molecular_weight" can be described as "Molecular weight in g/mol" and this is it. 60 characters was way too long for some of the simple columns.

PR #1500, opened 2025-01-10
py3.12 incompatibility with scipy and rdkit update

### LINKED ISSUE #1494: 🐛 Bug: Ersilia Test on PY3.12
### Describe the bug.

I am trying to use the test command at Ersilia. I have tried to install it with
`pip install -e .[test]`
In an environment that has Python 3.12, as in principle Py3.12 is supported by Ersilia. I have run into the following errors:

1. The rdkit-pypi I think is downgraded, I get the following error that disappears if I change lines 63 and 73 of the `pyproject.toml` file:
```
INFO: pip is looking at multiple versions of ersilia to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement rdkit-pypi (from ersilia) (from versions: none)
ERROR: No matching distribution found for rdkit-pypi
```
2. Once this is bypassed I get stuck at this issue (attached .txt). 
[ersiliatest_py312.txt](https://github.com/user-attachments/files/18345356/ersiliatest_py312.txt)

I believe this is due to Python 3.12 because in an environment with 3.11 it works. Also, The Ersilia install without [test] seems to work in Py3.12. Tagging both @Abellegese as the developer of the testing and @DhanshreeA 

_No response_

### Operating environment

Ubuntu 24.04 LTS

Abellegese:
**Description**
- [x] rdkit removed from the test extra in the `pyproject.toml` and added as fixture in the `test_run.py`, to be installed by Ersilia  when necessary.
- [x] scipy version version specified in  `pyproject.toml` `version = "<=1.10.0"` is not supported by  `py3.12` and updated to be version = "*` to be compatible withit.

Fixes #1494

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,6 @@ fuzzywuzzy = { version = "^0.18.0", optional = true }
 sphinx = { version = ">=6.0.0", optional = true }           # for minimum version and support for Python 3.10
 jinja2 = { version = "^3.1.2", optional = true }
 scipy = { version = "<=1.10.0", optional = true }
-rdkit-pypi = { version = "*", optional = true }
 nox = { version = "*", optional = true }
 rich = { version = "*", optional = true }
 ruff = { version = "*", optional = true }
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -59,8 +59,7 @@ pytest-benchmark = { version = "<=4.0.0", optional = true }
 fuzzywuzzy = { version = "^0.18.0", optional = true }
 sphinx = { version = ">=6.0.0", optional = true }           # for minimum version and support for Python 3.10
 jinja2 = { version = "^3.1.2", optional = true }
-scipy = { version = "<=1.10.0", optional = true }
-rdkit-pypi = { version = "*", optional = true }
+scipy = { version = "*", optional = true }
```
DhanshreeA, line 62: Why did we change this scipy version?

DhanshreeA (changes requested):

DhanshreeA:
@Abellegese LGTM mostly.

Abellegese:
Okay @DhanshreeA the reason I put the scipy version * was the current version does not supported by py3.12, so optional makes it more flexible installation for a given ersilia environment. It works.

DhanshreeA (approved):

DhanshreeA:
I can confirm that I was both able to reproduce the issue that @GemmaTuron had flagged, which is resolved through this PR.

Issue #1484, opened 2025-01-03
🐕 Batch: Refactoring Test workflows in models


DhanshreeA:
### Summary

This issue will encompass efforts to reconcile, clean up, and enhance our test (and build) pipelines for individual models. 

We currently have a [`test module`](https://github.com/ersilia-os/ersilia/blob/master/ersilia/publish/test.py) and CLI command (`ersilia test ...`) that can check a given model for functionality, completeness, and correctness. In addition to this, we also have a [testing playground](https://github.com/ersilia-os/ersilia/tree/master/test/playground) - a test utility which checks a given model for functionality, completeness, and correctness; and is able to simulate running one or more models on a user's system. 

[Existing](https://github.com/ersilia-os/eos-template/blob/950bd700ab69eff3e8db6c41015b86920dbbde04/.github/workflows/test-model.yml#L138) [test](https://github.com/ersilia-os/eos-template/blob/950bd700ab69eff3e8db6c41015b86920dbbde04/.github/workflows/upload-ersilia-pack.yml#L89) in our model [pipeline](https://github.com/ersilia-os/eos-template/blob/950bd700ab69eff3e8db6c41015b86920dbbde04/.github/workflows/upload-bentoml.yml#L109) is quite redundant in face of these functionalities because it is quite naive in comparison since it only tests nullity in  model predictions, and is [not robust](https://github.com/ersilia-os/eos-template/issues/83) to how a model might serialize its outputs. Moreover, the Docker build pipelines are bloated with code that can be removed in favor of a singular workflow testing the built images. We also need to handle testing for ARM and AMD builds more smartly since currently we only test the AMD images, but recently we have experienced [some](https://github.com/ersilia-os/eos8ub5) [models](https://github.com/ersilia-os/eos2db3/issues/1) successfully building for the ARM platform but then not actually working. 

Furthermore, we need to revisit [H5 serialization](https://github.com/ersilia-os/ersilia/issues/1479) within Ersilia, and also include tests for this functionality at the level of testing models.

Each of the objectives below should be considered individual tasks, and should be addressed in separate PRs referencing this issue.

### Objective(s)

- [x] Consolidate the following input-output combinations in the testing scenarios covered by the ersilia test command:

1. Input = CSV - Output = CSV
2. Input = CSV - Output = HDF5
3. Input = CSV - Output = JSON
4. Input = SMILE - Output = CSV
5. Input = SMILE - Output = HDF5
6. Input = SMILE - Output = JSON

[...]

PR #1514, opened 2025-01-16
Airtable field update and add unit testing

Abellegese:
#### Overview:
This PR modifies the aitable interface and `BaseInformation` class to incorporate the new fields as below:

```yml
Environment Size # to dependency size
Image Size # model docker image size
Computational Performance 1 # computational performance for 1 input sample
Computational Performance 10 # computational performance for 10 input samples
Computational Performance 100 # computational performance for 100 input samples
```
 Also to add adds unit tests for the `ReadmeMetadata`, `BaseInformation` classes and Mocked Airtable responses are used for tests.

- **`ReadmeMetadata`**: Tests the `write_information()` methods to ensure correct data retrieval and README generation.
- **`BaseInformation`**: Verifies that model attributes (title, description, performance metrics, etc.) are correctly set and retrieved using direct assignment.

#### Tests Added:
- **`test_read_information`**: Confirms correct model data retrieval from Airtable.
- **`test_write_information`**: Ensures the README content is formatted correctly.
- **`test_write_information_to_file`**: Verifies that README content is written to a file.
- **`test_base_information_setter`**: Tests direct assignment of attributes in `BaseInformation`.
- **`test_airtable_interface_ro`**: Verifies Airtable read-only functionality.
- **`test_airtable_interface_rw`**: Verifies Airtable read-write functionality.
- **`test_airtable_metadata_find_record`**: Confirms correct data retrieval from Airtable metadata.

Related to #1484

### DIFF
--- a/.github/workflows/tests_and_cleanup.yml
+++ b/.github/workflows/tests_and_cleanup.yml
@@ -102,7 +102,7 @@ jobs:
 
       - name: Run pytest
         run: |     
-          pip install pytest pytest-benchmark pytest-asyncio rdkit-pypi nox rich fuzzywuzzy scipy
+          pip install pytest pytest-benchmark pytest-asyncio rdkit-pypi nox rich fuzzywuzzy scipy pyairtable
           pytest
 
   run-cli-test-single:
--- a/ersilia/hub/content/base_information.py
+++ b/ersilia/hub/content/base_information.py
@@ -73,7 +73,7 @@ class BaseInformation(ErsiliaBase):
         Attributes
         ----------
         _github : None
-            Placeholder for GitHub-related data.
+            Placeholder for GitHub URL of the model.
         _identifier : None
             Placeholder for a unique identifier string.
         _slug : None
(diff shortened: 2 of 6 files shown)


.github/workflows/tests_and_cleanup.yml
```diff
@@ -102,7 +102,7 @@ jobs:
 
       - name: Run pytest
         run: |     
-          pip install pytest pytest-benchmark pytest-asyncio rdkit-pypi nox rich fuzzywuzzy scipy
+          pip install pytest pytest-benchmark pytest-asyncio rdkit-pypi nox rich fuzzywuzzy scipy pyairtable
```
DhanshreeA, line 105: Why do we need airtable as a dependency for the tests?

ersilia/hub/content/base_information.py
```diff
@@ -66,6 +142,11 @@ def __init__(self, config_json):
         self._description = None
         self._mode = None
         self._task = None
+        self._subtask = None
```
DhanshreeA, line 145: I am not sure if we need to add this?

ersilia/hub/content/base_information.py
```diff
@@ -83,6 +164,13 @@ def __init__(self, config_json):
         self._docker_architecture = None
         self._s3 = None
         self._memory_gb = None
+        self._source = None
```
DhanshreeA, line 167: What are `_source`, and `_source_type` for? I don't understand where in Airtable do they fit?

DhanshreeA (changes requested):

.github/workflows/tests_and_cleanup.yml
Abellegese: I mean there was a code in `airtableops.py` that imports `pyairtab;e` which first used here

```python
class AirtableInterface:
    def __init__(self, mode="ro", api_key=None):
        self.base_id = AIRTABLE_MODEL_HUB_BASE_ID
        self.table_name = AIRTABLE_MODEL_HUB_TABLE_NAME
...
    def _create_table(self, api_key):
        return pyairtable.Table(api_key, self.base_id, self.table_name)
```

ersilia/hub/content/base_information.py
Abellegese: I they were a new added fields in the Airtable. I first found them when I refactor the `ruff` and in this PR I added them in the in `as_dict`, in `BaseInformation` class.

Abellegese: There are a few other more fields as well.

DhanshreeA:
I will update the docstring with more helpful explanation of the fields, but otherwise the PR is great!

Abellegese:
Thanks @DhanshreeA

DhanshreeA:
Update, I've removed the `Memory GB` field from the `BaseInformation` class, since it is redundant and nowhere being used. We should drop this from AirTable as well, as @miquelduranfrigola had pointed out.

Issue #1509, opened 2025-01-15
🐈 Task: Revise resolution method for Dockerized images in models


DhanshreeA:
### Summary

Presently we determine the packing method of a model's image using this utility [function](https://github.com/ersilia-os/ersilia/blob/5dc3e30f51e3dc93e5d88c0c0f7568f843633e46/ersilia/utils/docker.py#L27), which looks at the history of the docker image and determines how the image was created based on the kinds of commands that were run in the image layers during its creation. 

This is turning out to be an unreliable solution as on some users' machines, the image's history appears to be different, despite the image being entirely the same (as we can verify from its SHA256 value). Therefore, we have decided to add another metadata field to models called `Docker Pack Method` which we intend to utilise instead of reading the image's history to resolve the image's pack method. 

This information can be easily collected and stored in our metadata stores as part of our workflows. And utilising this would make ersilia a lot more reliable and not cause perfectly working models to randomly break every now and then.

### Objective(s)

- [x] Update the `Docker Pack Method` column in the Airtable database for all models.
- [x] Update the referenced function to read the metadata. In fact the function can be deleted, and we can potentially add a method in the `BaseInformation` class.
- [x] Update the model build workflows to capture this information.
- [x] Update `airtableops.py` module to add this information into Airtable, which will then be naturally serialized to the S3 JSON as well.

DhanshreeA:
@GemmaTuron this should take care of what you're experiencing here: https://github.com/ersilia-os/eos3b5e/issues/15#issuecomment-2382571783

DhanshreeA:
Part of updating the workflows will be addressed in the #1484

GemmaTuron:
Hello @DhanshreeA 

At present I cannot run models in my local computer because they consistently run into this problem. The resolver decides it is a bentoML packed model when it is not, and hence it ends up crashing:

```
Error message:

Ersilia exception class:
StandardModelExampleError

Detailed error:
Standard model run from CSV was not possible for model eos3b5e
Output file /home/gturon/eos/dest/eos3b5e/example_standard_output.csv was not created successfully

Hints:
If you fetch this model from Docker Hub, or you are running it through URL, this is the first time run is executed in your local computer. Reach out to Ersilia to get specific help.
```
I see this task is on hold but to me it is prioritary to solve.

GemmaTuron:
Okay,

[...]

PR #1533, opened 2025-02-04
fix resolve pack method

GemmaTuron:

**Description**

The resolve pack method for Docker images was not working as the history had several versions some of which contained nginx, which is what was defined to decide between bentoml or fastapi

**Changes to be made**

I have added an if statement to check only in the latest version

**Status**

Completed. I have tried to use it with FastAPI packed models that were originally packed with BentoML (eos3b5e) and 
models that have been packed with BentoML (eos9f6t) for example. Works in both situations.

**To do**

None

Related to #1509

### DIFF
--- a/ersilia/utils/docker.py
+++ b/ersilia/utils/docker.py
@@ -31,9 +31,11 @@ def resolve_pack_method_docker(model_id):
     model_image = client.images.get(f"{DOCKERHUB_ORG}/{model_id}:{docker_tag}")
     image_history = model_image.history()
     for hist in image_history:
-        # Very hacky, but works bec we don't have nginx in ersilia-pack images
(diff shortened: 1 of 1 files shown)


GemmaTuron:
Okay definitely looking into this history is no good way of picking the packaging method

I tried in another computer and I got an absolute different output for the history of eos3b5e, which does not contain the "nginx" string anywhere not even in the not "latest" images. I am guessing cache plays a role here

GemmaTuron:
okay definitely according to GPT the history of an image might differ between the python API and the CLI. Why the behavior is different between computers is a mistery and probably is due to a cached file that does not get removed

@Abellegese don't merge the PR as this will not fix the issue as it won't fix the issue system wide probably

Issue #1538, opened 2025-02-06
🐕 Batch: Model Test and Upgrade


GemmaTuron:
### Summary

The last six months have had a good deal of new developments in Ersilia. We need to consolidate these changes and ensure all the models work and remain working. There are several moving pieces that will need to be coordinated in a sprint this February.

### Objective(s)

As a general overview, these are the main objectives that will need to be tackled and in which order. Each Objective is linked to an Epic and each Epic will have a series of tasks to be completed.
1. Fix critical bugs in Ersilia (i.e those bugs that render models not working at the Ersilia level, not individual level) - https://github.com/ersilia-os/ersilia/issues/1536
In paral·lel:
2. Complete the development of the Test Module v1, including if possible tests for (1) - https://github.com/ersilia-os/ersilia/issues/1484
2. Upgrade the model metadata to the newest version in Airtable & synchronise with all repos - https://github.com/ersilia-os/ersilia/issues/1477
Next:
3. Revise the test Module once the new metadata is updated, and ensure the changes are accounted for: https://github.com/ersilia-os/ersilia/issues/1537
4. Upgrade the workflows in eos-template to use the test module at model push: https://github.com/ersilia-os/eos-template/issues/90
5. Fix bugs / Improve the FastAPI packaging: I am less expert to comment on those, I'll refer to the three open issues in the ersilia-pack repo: https://github.com/ersilia-os/ersilia-pack/issues/49, https://github.com/ersilia-os/ersilia-pack/issues/48, https://github.com/ersilia-os/ersilia-pack/issues/46
6. Upgrade models in batches of 10 to the newest FastAPI packaging and ensure they pass the tests: we will create another batch for this probably

GemmaTuron:
We have tackled points 1,2, 4 and 6

5 is almost complete and then we'll go into batch completing 3 and 7

GemmaTuron:
We are actively working on 7. All other tasks are completed, I'll close this Batch.

PR #1572, opened 2025-02-26
fix BI exceptions, Create class to update metadata from airtable, rea…

GemmaTuron:
…nnotation-json for S3

**Description**

This branch contains a version of Ersilia that works only with the new metadata fields

**Changes to be made**

Things to consider in the new branch:

- The metadata fields not present in airtable anymore have been eliminated. This is necessary @miquelduranfrigola  because otherwise it will complain that those fields are not in airtable, I think. Backwards compatibility is very difficult. We can potentially reinstate all fields in airtable, and THEN keep everything in the first round and eliminate it in a second round.
- The MODELS_JSON now points to a models-reannotated.json (which does not exist, yet) to avoid overwriting the models.json we currently have. The script convert_airtable_to_json also points to this new file, so it could be created by simply running this workflow
- The pyairtable.Table() has been updated to the new version using the Api module (this was deprecated since py 3.9)
- The airtableops.py script now has a metadata-update option that will update the metadata file in the repository from the information in the Reannotated Airtable. I have tested this in local and it works. We should run this as the first step in the workflows, only once, then delete it.
- The airtableops.py does not yet parse the new information into a README appropriately. This is the work @OlawumiSalaam  has been doing. We can synch on this.

**Status**
In progress
**To do**
Coordinate on the model workflow updates, if we want them to use this branch even if not merged in main

Related to #1538


GemmaTuron:
In this branch we currently have:

- All metadata fields possible (old and new) in the Base Information
- Automated resolution to None for some fields required by Ersilia (like Input Shape) and subsequent handling appropriately
- New metadata updater from airtable
- New readme updater (to be tested) for the new metadata

To DO:

- Test more extensively before merging
- Ensure the readme updater works and modify the card.py file
- Decide where / when to read from which Airtable (currently set to the new airtable)
- Make a strategy for model upload

GemmaTuron:
Updates:

- [ ] Worked on the README formatting and ensuring the right version gets called when necessary
- [ ] Adapted the test_base_information on pytests to check for the new readme file format
- [ ] Fixed small errors in airtableops and made the ro key work for the new airtable as well

Next:
- [ ] Test this branch for model fetching and serving thoroughly
- [ ] Decide how to merge this branch so that the workflows start to run the new version of Ersilia

Proposed plan:
Example models:
- eos3b5e (modify to old python version to have a BentoML model)
- eos5axz (old format file but packaging with FastAPI)
- eos74km (new format file)

Clone these models locally and test fetching from_dir with the new and the old metadata formats. To update the metadata, use the metadata reader in airtableops with local pat key.
@miquelduranfrigola how could we test thoroughly from_github and from_dockerhub when we do not have the latest images available?

PR #1672, opened 2025-06-12
Model incorporation debugging

miquelduranfrigola:
This pull request includes a minor change to the `_fetch_schema_from_github` method in the `ersilia/io/output.py` file. The change removes an unnecessary condition in the `if` statement that checks for an empty `shape` variable.

* Simplified condition in `_fetch_schema_from_github`: Removed the redundant check for `shape == 0` in the `if` statement, as it is not necessary for determining whether to return `None`. (`[ersilia/io/output.pyL437-R437](diffhunk://#diff-d045a06cd71f476160524f7bc056c76f761e3af066c00fbd63fd13d8a264171aL437-R437)`)

### DIFF
--- a/ersilia/hub/content/card.py
+++ b/ersilia/hub/content/card.py
@@ -173,8 +173,9 @@ class MetadataCard(ErsiliaBase):
 
     def __init__(self, config_json):
         ErsiliaBase.__init__(self, config_json=config_json)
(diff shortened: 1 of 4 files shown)


miquelduranfrigola:
@Abellegese very importantly, there is a message addressed to you in the standard_api.py script, related to parametrization of the request.

Issue #1839, opened 2026-04-16
🐛 Bug: Ersilia stops execution if a batch fails


Marina18:
### Describe the bug.

If a molecule in a batch could not be processed ersilia stops the execution

### Operating environment

Ubuntu24.0

PR #1841, opened 2026-04-23
feat: divide-and-conquer fallback for failing molecules in batch runs

Marina18:

**Description**

This PR addresses two issues: (1) a crash when running models where one or more molecules in a batch cause the API to error, previously the whole run failed with no output; (2) a `TypeError` crash when serving a model that was fetched from DockerHub and has no local bundle directory.

**Changes to be made**

- Add` _post_batch_with_fallback` to `StandardCSVRunApi`: when a batch POST fails, split it in half and retry each half recursively (divide and conquer). Single molecules that consistently fail are assigned empty output columns in the final CSV rather than aborting the run. Input/output row order is preserved.
- Remove the automatic printing of the full output CSV/JSON content to the terminal after `ersilia run` completes.
- Guard` _get_bundle_location()` returning None in `autoservice.py` and `base.py` so that serving Docker-fetched models  no longer raises a `TypeError`.
- Change `_LocalService.__init__` to set `self.server = None` instead of raising an exception when the pack method is not FastAPI, allowing `is_available()` to be called safely during service class auto-detection.

**Status**
Implemented and tested locally against a Docker-served model with a mixed input containing valid and invalid molecules. The output CSV correctly contains results for all valid molecules and empty columns for failing ones, and the run no longer crashes.

Related to #1839

### DIFF
--- a/ersilia/cli/commands/run.py
+++ b/ersilia/cli/commands/run.py
@@ -100,11 +100,5 @@ def run_cmd():
             fg="green",
             bold=False,
         )
-        if output and output.endswith(".csv"):
-            with open(output, "r") as f:
-                echo(f.read())
-        elif output and output.endswith(".json"):
-            with open(output, "r") as f:
-                echo(json.dumps(json.load(f), indent=2))
 
     return run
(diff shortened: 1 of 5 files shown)


miquelduranfrigola:
@Marina18 I have had a quick look at the code, but not tested it. It looks good. In my opinion, this is ready to merge.

PR #1878, opened 2026-06-21
Update ersilia API

Lehcar:

## Description

This PR updates the ersilia API (create_api.py) to:
- Check if docker is running when fetching models if no source is defined
- Return True if fetch/close/delete succeeds (and False if fetch/close/delete fails)
- Return information on where the model is served if serve() succeeds (it'll throw a RuntimeError if it fails to serve)
- Docs updates (including adding information on running tests locally) 

The linter (`ruff format`) also made changes to these files, but the functionality should be the same:
- `ersilia/cli/commands/__init__.py`
- `ersilia/db/environments/managers.py`
- `ersilia/hub/content/base_information_validator.py`
- `ersilia/hub/delete/delete.py`
- `ersilia/utils/docker.py`

N/A

------------------
## Testing
Ran `pytest -v`
image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,7 @@ eosvc = { version = ">=1.1.0", optional = true }
 pytest = { version = "^7.4.0", optional = true }
 pytest-asyncio = { version = "<=0.24.0", optional = true }
 pytest-benchmark = { version = "<=4.0.0", optional = true }
+pyairtable = { version = "*", optional = true }
 fuzzywuzzy = { version = "^0.18.0", optional = true }
 sphinx = { version = ">=6.0.0", optional = true }           # for minimum version and support for Python 3.10
 jinja2 = { version = "^3.1.2", optional = true }
@@ -75,7 +76,7 @@ isaura = { version = ">=2.2.0", optional = true }
 docs = ["sphinx", "jinja2"]
 isaura = ["isaura"]
 eosvc = ["eosvc"]
-test = ["pytest", "pytest-asyncio", "pytest-benchmark", "nox", "rich", "fuzzywuzzy", "scipy", "ruff", "pre-commit"]
+test = ["pytest", "pytest-asyncio", "pytest-benchmark", "pyairtable", "nox", "rich", "fuzzywuzzy", "scipy", "ruff", "pre-commit"]
 #all = [docs, test]
 
 [tool.poetry.scripts]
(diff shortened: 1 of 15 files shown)


pyproject.toml
```diff
@@ -59,6 +59,7 @@ eosvc = { version = ">=1.1.0", optional = true }
 pytest = { version = "^7.4.0", optional = true }
 pytest-asyncio = { version = "<=0.24.0", optional = true }
 pytest-benchmark = { version = "<=4.0.0", optional = true }
+pyairtable = { version = "*", optional = true }
```
Lehcar: I added this here because I got the following error when I tried running pytest locally:
```bash
ModuleNotFoundError: No module named 'pyairtable
```

It looks like pyairtable is installed separately in Ersilia's GitHub Actions workflows ([example 1](https://github.com/ersilia-os/ersilia/blob/dfd8cef9064709a304f1eb5009cef4136a6a4ff5/.github/workflows/approve-dispatch.yml#L181), [example 2](https://github.com/ersilia-os/ersilia/blob/dfd8cef9064709a304f1eb5009cef4136a6a4ff5/.github/workflows/tests_and_cleanup.yml#L51)) so I'm not sure if pyairtable was intentionally left out of pyproject.toml. If that was intentional, I can undo my changes to pyproject.toml and update the development docs 👍

test/api/test_model.py
```diff
@@ -6,20 +8,63 @@
     "CC1C2C(CC3(C=CC(=O)C(=C3C2OC1=O)C)C)O"
 ]
 
-def test_fetch():
+@patch('ersilia.api.create_api.Model._is_docker_running')
```
Lehcar: I also want to highlight that these tests aren't really testing the underlying api operations anymore because they're written more like unit tests rather than integration tests (I couldn't get them to pass without docker running locally on my system).

If we want these to be integration tests again, I can either:
1. Rewrite this so we're actually testing the underlying operations and I can update DEVELOPMENT.md to note that docker needs to be running in order to run these tests, or
2. Add a separate integration test file and ensure those are ran by the GitHub Action

2 would be better if Docker isn't a hard requirement, otherwise I'm thinking 1 is better

Lehcar: I thought about this and I'm going to update these tests so they're integration tests again and update the development docs

pyproject.toml
Lehcar: Discussed this over slack. The pyairtable dependency has been unstable in the past, so we do not want to include it in the base ersilia installation. I'll undo this change and update the docs to note to install this separately to run tests locally

Lehcar: Done in https://github.com/ersilia-os/ersilia/pull/1878/commits/f245a12e74d3a52c0a222ff4daaf101e0685eee8

miquelduranfrigola:
@Marina18 can you double-check this and merge if it looks good (it does to me).

Marina18 (approved):

PR #1905, opened 2026-08-15
Fix TypeError in metadata file selection for metadata.yml-only repos in remote mode

### LINKED ISSUE #1902: airtableops.py: _select_correct_metadata_file raises TypeError for metadata.yml-only repos in remote mode
## Summary

`MetadataFileUpdater._select_correct_metadata_file` in `.github/scripts/airtableops.py` raises a `TypeError` in remote mode for any model repository whose only metadata file is `metadata.yml`.

```python
def _select_correct_metadata_file(self):
    if self.repo_path is None:                                                   # remote mode
        if os.path.exists(os.path.join(self.tmp_folder, self.model_id, "metadata.json")):
            return os.path.join(self.tmp_folder, self.model_id, "metadata.json")
        elif os.path.exists(os.path.join(self.tmp_folder, self.model_id, "metadata.yaml")):
            return os.path.join(self.tmp_folder, self.model_id, "metadata.yaml")
        elif os.path.exists(os.path.join(self.repo_path, "metadata.yml")):       # repo_path is None here
            return os.path.join(self.repo_path, "metadata.yml")
```

Two problems in the `repo_path is None` branch:

1. The third condition uses `self.repo_path`, which is `None` in this branch, instead of `self.tmp_folder` / `self.model_id`. It raises `TypeError: expected str, bytes or os.PathLike object, not NoneType`.
2. The second condition looks for `metadata.yaml`, but model repositories use `metadata.yml`.

Together these mean the remote branch can never reach a `.yml` file: the `.json` check misses, the `.yaml` check misses, and the `.yml` check raises before it can match.

## Reproduction

```
python airtableops.py metadata-update --repo <model_id> --api-key <key>
```

for any model whose repository contains `metadata.yml` and no `metadata.json`.

## Impact

`metadata-update` is the Airtable to repository direction, so this blocks writing metadata back to those repositories. Of the 218 Ready models, **87 have `metadata.yml` as their only metadata file** and would hit this. Passing `--path` with a locally cloned repository takes the `else` branch, which is correct, so the bug only affects remote mode.

## Suggested fix

In the `repo_path is None` branch, resolve all three candidates against `os.path.join(self.tmp_folder, self.model_id, ...)` and include `metadata.yml` alongside `metadata.yaml`.

## Context

Found while assessing whether the Airtable to repository sweep could be used to apply metadata corrections from a hub-wide metadata audit.

SpyderMYK:
Fixes #1902

`MetadataFileUpdater._select_correct_metadata_file` resolved its third candidate against `self.repo_path` inside the `repo_path is None` branch, raising `TypeError` in remote mode for any model repo whose only metadata file is `metadata.yml`; the second candidate also checked `metadata.yaml` where model repos use `metadata.yml`. This collapses the two duplicated branches into one probe over `("metadata.json", "metadata.yaml", "metadata.yml")` against the correct base directory, preserving probe order and the existing not-found behavior.

Also fixed as a side effect: in remote mode with no metadata file at all, the function previously raised the same `TypeError` instead of reaching the "Metadata file not found" message.

Verified with a four-case matrix (remote/.yml-only, remote/json-preference, local/.yml, remote/none): the first and last cases fail on master with `TypeError` and pass with this change; the middle two pass before and after. Pre-existing black violations elsewhere in the file were left untouched to keep the diff scoped.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/scripts/airtableops.py
+++ b/.github/scripts/airtableops.py
@@ -220,23 +220,14 @@ class MetadataFileUpdater(FileUpdater):
 
     def _select_correct_metadata_file(self):
         if self.repo_path is None:
-            if os.path.exists(os.path.join(self.tmp_folder, self.model_id, "metadata.json")):
-                return os.path.join(self.tmp_folder, self.model_id, "metadata.json")
-            elif os.path.exists(os.path.join(self.tmp_folder, self.model_id, "metadata.yaml")):
-                return os.path.join(self.tmp_folder, self.model_id, "metadata.yaml")
-            elif os.path.exists(os.path.join(self.repo_path, "metadata.yml")):
-                return os.path.join(self.repo_path, "metadata.yml")
-            else:
-                print("Metadata file not found")
+            base_dir = os.path.join(self.tmp_folder, self.model_id)
         else:
-            if os.path.exists(os.path.join(self.repo_path, "metadata.json")):
-                return os.path.join(self.repo_path, "metadata.json")
-            elif os.path.exists(os.path.join(self.repo_path, "metadata.yaml")):
-                return os.path.join(self.repo_path, "metadata.yaml")
-            elif os.path.exists(os.path.join(self.repo_path, "metadata.yml")):
-                return os.path.join(self.repo_path, "metadata.yml")
-            else:
-                print("Metadata file not found")
+            base_dir = self.repo_path
+        for filename in ("metadata.json", "metadata.yaml", "metadata.yml"):
+            path = os.path.join(base_dir, filename)
+            if os.path.exists(path):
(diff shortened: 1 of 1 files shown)


SpyderMYK:
Hi @GemmaTuron — this PR fixes the `TypeError` you reported in #1902.

As a first-time contributor, my workflow runs need maintainer approval before they'll start, so only the GitGuardian check has run so far (passing). Whenever you get a chance, could someone approve the CI workflows so the test suite can run against the change? The diff is small — one method in `.github/scripts/airtableops.py`.

Thanks!

